Issue #213, opened 2023-08-16
Better error when no app is found


ddieruf:
I had configured the cli to use tenant-1 but was working on an app in tenant-2. When I ran the following:

`./bin/sga-cli gateway produce text-processing simple-produce --value "Hi there"`

I got:

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
The HTTP response from the server [500] did not permit the HTTP upgrade to WebSocket
```

When I looked at the gateway logs I discovered that:

`java.lang.IllegalArgumentException: application text-processing not found`

Because I had the wrong tenant set. It would have helped to receive a better message from the CLI.

PR #236, opened 2023-08-21
Improve CLI gateway error messages

nicoloboschi:
Fixes: #213

* Perform preliminary checks for the gateway when using the CLI. Unfortunately there's no way to get the server error during the handshake so I reimplemented the same checks. Refactored the base command to use a new LangStream HTTP java client that could be reused in future as java admin sdk.
* Fix applications updates to use PATCH as expected now by the control plane
* Now all the CLI commands accept the `--help` flag
* Added secrets file validation on control plane and fixed example


eolivelli (approved):
LGTM

PR #242, opened 2023-08-21
Update README.md to include google client id

mendonk:
To run the trial example, it needs google client id or you get an auth error. You can alternatively comment out the produce and consume topics in the gateway yaml to not use this ID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ In this example, you will deploy an application that performs AI completion and 
 
 OPEN_AI_URL=xx
 OPEN_AI_ACCESS_KEY=xx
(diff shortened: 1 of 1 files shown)


nicoloboschi (changes requested):
@mendonk in that example you have both authenticated and unauthenticad gateways. The one suggested by the readme is without authentication. 
Adding the google client id only slows down the user to get their first app deployed so I think it's better to not commit this change

eolivelli:
the main README is only for a quick trial, I wouldn't add Google Authentication for Gateways, that is an advanced topic

mendonk:
Closing this and raising an issue

PR #244, opened 2023-08-22
Move docker images to Github Container Registry

nicoloboschi:
* Release process now pushes to ghcr.io/LangStream/langstream-XX
* Docker images generated locally will be under langstream instead of datastax. e.g. datastax/langstream-runtime -> langstream/langstream-runtime


eolivelli (approved):
LGTM

PR #245, opened 2023-08-22
Deployer: set infinite retries

nicoloboschi:
* Currently the deployer uses the default retry config. This retry config is applied when the reconciliation loop fails. The correct behaviour would to be retry forever until the patch is in the requested state otherwise it seems everything is fine even if it's not able to update the resources for some reasons.


eolivelli (approved):
LGTM

PR #250, opened 2023-08-22
Add cassio and llama-index deps

cbornet:


eolivelli (approved):
LGTM

looking forward to see some examples

cc @erichare FYI

PR #252, opened 2023-08-23
Package system agents with isolated classpath (using NiFi Nar)

eolivelli:
Summary:
- in order to prevent classpath pollution we have to package each system agent into a self-contained bundle
- at runtime we load each agent from their own package

### DIFF
--- a/langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
+++ b/langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
@@ -16,6 +16,7 @@
 package ai.langstream.runtime.agent;
 
 import ai.langstream.api.runner.code.AgentCode;
+import ai.langstream.api.runner.code.AgentCodeAndLoader;
 import ai.langstream.api.runner.code.AgentCodeRegistry;
 import ai.langstream.api.runner.code.AgentContext;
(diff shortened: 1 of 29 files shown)


langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
```diff
@@ -140,28 +146,29 @@ public static void run(RuntimePodConfiguration configuration,
 
         log.info("Starting agent {} with configuration {}", agentId, configuration.agent());
 
-        TopicConnectionsRuntime topicConnectionsRuntime =
-                TOPIC_CONNECTIONS_REGISTRY.getTopicConnectionsRuntime(configuration.streamingCluster());
+        try (NarFileHandler narFileHandler = new NarFileHandler(agentsDirectory);) {
```
cbornet, line 149: So now we only support loading NARs ? Or is it possible to also load JARs being aware that we lose isolation ?

cbornet (approved):

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
eolivelli: this is for internal agents only at the moment, at some point we could add custom Java agents
you can already bundle custom jars (for Kafka Connect and for JDBC)

PR #255, opened 2023-08-23
Create Java Admin client module

nicoloboschi:
This basically split the CLI code and the actual http client for the control plane. 
For now only applications are implemented. 
This is a preparatory work for moving the code download to use the new endpoint instead of s3 directly


eolivelli (approved):
+1

PR #257, opened 2023-08-23
Add CLI autocomplete instructions for Bash

### LINKED ISSUE #256: Give precisions on the auto-completion setup
In the README, we give instructions to setup auto-completion: https://github.com/LangStream/langstream#enable-auto-completion
This is only for ZSH. The user may use another shell and this won't work.
We need to be clear that this is for ZSH and provide instructions for other common shells if possible

nicoloboschi:
Fixes #256

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,20 @@ There are multiple ways to install the CLI.
   ```  
 
 ### Enable auto-completion
-Installing directly the binary will enable auto-completion for the CLI. If you installed the CLI with Homebrew, you can enable it with the following command:
+Installing directly the binary will enable auto-completion for the CLI. 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,10 +87,20 @@ There are multiple ways to install the CLI.
   ```  
 
 ### Enable auto-completion
-Installing directly the binary will enable auto-completion for the CLI. If you installed the CLI with Homebrew, you can enable it with the following command:
+Installing directly the binary will enable auto-completion for the CLI. 
+
+If you installed the CLI with Homebrew, you can enable it with the following command:
```
cbornet, line 92:
```suggestion
If you installed the CLI with Homebrew, you can enable auto-completion with the following command:
```

PR #263, opened 2023-08-24
Move runtime pod to download code using the control plane 

nicoloboschi:


langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentCodeDownloader.java
```diff
@@ -16,62 +16,69 @@
...
+            }
 
+            @Override
+            public void error(Object message) {
+                log.error(message.toString());
 
-        if (codeStorageConfig != null) {
-            log.info("Downloading custom code from {}", codeStorageConfig);
-            log.info("Custom code is stored in {}", codeDownloadPath);
-            try (CodeStorage codeStorage =
```
eolivelli, line 68: I wonder if we could have kept this new code as simply a new implenetation of CodeStorage.
ControlPlaneCodeStorage,
now we are binding strictly the download of the code to the control plane and we don't have anymore this abstraction here.

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentCodeDownloaderStarter.java
```diff
@@ -0,0 +1,97 @@
...
+        final Path tokenPath = getPathFromEnv(AgentCodeDownloaderConstants.TOKEN_ENV,
+                AgentCodeDownloaderConstants.TOKEN_ENV_DEFAULT);
+
+        DownloadAgentCodeConfiguration configuration = MAPPER.readValue(downloadCodeConfigPath.toFile(),
+                DownloadAgentCodeConfiguration.class);
+
+        ClusterConfiguration clusterConfiguration = MAPPER.readValue(clusterConfigPath.toFile(),
+                ClusterConfiguration.class);
+        final String token;
+        if (tokenPath.toFile().exists()) {
```
eolivelli, line 73: nit: Files.isRegularFile

langstream-runtime/langstream-runtime-impl/src/test/java/ai/langstream/AbstractApplicationRunner.java
```diff
@@ -123,7 +123,7 @@ protected String buildInstanceYaml() {
 
     @BeforeAll
     public static void setup() throws Exception {
-         applicationDeployer = ApplicationDeployer
+        applicationDeployer = ApplicationDeployer
```
eolivelli, line 126: the changes in this file are no related ?

langstream-runtime/langstream-runtime-impl/src/test/java/ai/langstream/runtime/agent/AgentCodeDownloaderStarterTest.java
```diff
@@ -0,0 +1,87 @@
...
+    private void runTest(boolean expectOk, Map<String, String> env, String... args) {
+        final AgentCodeDownloader codeDownloader = Mockito.mock(AgentCodeDownloader.class);
+
+        try {
+            new TestDeployer(env).run(codeDownloader, args);
+        } catch (Exception e) {
+            if (expectOk) {
+                throw new RuntimeException(e);
+            } else {
+                return;
```
eolivelli, line 63: should we perform assertions on the Exception ? here we can see anything else failing

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentCodeDownloader.java
nicoloboschi: I thought about that but the problem it's that the control plane download is not "generic". You need to read a secret for the control plane url, another secret for the token. This impacts both the deployer and the runtime. In the deployer you'd need to ask the code storage impl to customize the pod or load secrets or similar. It doesn't feel very generic. 

Additionally, this is generally more secure since the we know for sure that the control plane url is verified.

Another advantage is that you need to write the code storage configuration only in the control plane now, which facilitates upgrades

eolivelli (approved):
LGTM

PR #264, opened 2023-08-24
Python libs and PYTHONPATH cleanup

cbornet:


cbornet:
I had first reordered the deps so that the user could override the runtime deps.
But in the end, I think it's better to keep it as-is. So the order is:
* /app/python_libs (langstream and langstream_runtime)
* /app/app-code-/python (user deps)
* python deps (installed by pip when building the docker image)

So the user can override the python deps but not the langstream ones.

nicoloboschi (approved):

PR #266, opened 2023-08-24
Bump python runtime deps (llama_index to 0.8.8)

cbornet:
This updates llama_index to 0.8.8 which has support for cassio.


eolivelli (approved):
+1

PR #273, opened 2023-08-25
Add /app-code-download/python/lib to the PYTHONPATH

cbornet:
This way users can install target libs to their `python/lib` dir.
When they run locally they shouldn't have `python/lib` on their python path so the libs won't be used.


nicoloboschi (changes requested):
Can we add a unit test ? this is a critical code and we don't have any test
at very least verify the process has started with correct args/env

cbornet:
The most useful would be an e2e test. I think you're working on one ?

eolivelli (approved):
LGTM

eolivelli:
@cbornet there is already one end-to-end test with python

@nicoloboschi  I would prefer to commit the patch and unlock @ddieruf developer guide, but I am asking @cbornet to add the test.
We can simply enhance the existing Python test

nicoloboschi (approved):

PR #287, opened 2023-08-28
Fixes using OIDC on AWS

nicoloboschi:
* Added support for oidc issuer that doesn't resolve to local kubernetes dns
* Fixed certificate import at control plane startup to also load system certs
* Fixed authz check for admin accessing tenant's applications endpoints
* Added support for OIDC providers that exposes multiple keys (before we were getting the first one instead of the one in the jwt header)
* Agent runtime wasn't using the service account of that tenant, therefore it was authenticating to the control plane with the role "default" - my testing was done using the "default" tenant 😞


eolivelli (approved):
+1

PR #289, opened 2023-08-28
Add error handling in Python Sinks

cbornet:


langstream-runtime/langstream-runtime-impl/src/main/python/langstream/util.py
```diff
@@ -72,8 +72,7 @@ def process(self, records: List[Record]) -> List[Tuple[Record, Union[List[Record
         for record in records:
             try:
                 processed = self.process_record(record)
-                if len(processed) > 0:
-                    results.append((record, processed))
+                results.append((record, processed))
```
cbornet, line 75: I think there should always be a result for any input value

langstream-runtime/langstream-runtime-impl/src/main/python/langstream_runtime/runtime.py
```diff
@@ -145,7 +146,9 @@ def run_main_loop(source: Source, sink: Sink, processor: Processor, errors_handl
                                 # commit skipped records
                                 call_method_if_exists(source, 'commit', [source_record_and_result[0]])
                             else:
-                                sink.write(source_record_and_result[1])
+                                if len(source_record_and_result[1]) > 0:
```
cbornet, line 149: No need to call the sink if there's nothing to write

langstream-runtime/langstream-runtime-impl/src/main/python/langstream_runtime/source_record_tracker.py
```diff
@@ -39,14 +39,22 @@ def commit(self, sink_records: List[Record]):
...
         for source_record_and_result in sink_records:
             source_record = source_record_and_result[0]
             result_records = source_record_and_result[1]
             if not isinstance(result_records, Exception):
-                self.remaining_sink_records_for_source_record[source_record] = len(result_records)
-                for sink_record in result_records:
-                    self.sink_to_source_mapping[sink_record] = source_record
+                if len(result_records) == 0:
+                    # commit immediately if the result is empty
+                    source_records_to_commit.append(source_record)
```
cbornet, line 53: If there's no result returned by the processor then commit won't be called by the sink. So we need to commit right away.

langstream-runtime/langstream-runtime-impl/src/main/python/langstream/util.py
eolivelli: yes

eolivelli (approved):
+1

PR #294, opened 2023-08-29
Single command to manage local LangStream cluster 

nicoloboschi:


dev/llc.sh
```diff
@@ -0,0 +1,332 @@
...
+    label=langstream-client
+  fi
+  if [ ! -z "$label" ]; then
+    kubectl_cmd -n $k8s_namespace delete pod -l app.kubernetes.io/name=$label
+    echo "Pod $label: ✅"
+  fi
+}
+build_no_restart() {
+  eval $(minikube_cmd docker-env)
+  ./docker/build.sh "$@" &> /dev/null
```
eolivelli, line 295: please note that now docker/build.sh runtime is broken.
because now the runtime-impl module doen't depend anymore on the "agents", it is only copying the .nar files, so Maven with "-am" doesn't rebuild them.

in case of "runtime" we have to rebuild manually more stuff with "mvn clean install -DskipTests"

eolivelli: no need to fix it in this patch, it is unrelated

eolivelli (commented):
LGTM

As a matter of taste I don't like "llc", I prefer something more explicit, like
`dev/langstream-standalone`

eolivelli:
closing this for now

PR #307, opened 2023-08-30
Add topicName to Kafka consumer/producer agent info

cbornet:


cbornet:
Closing as this is already handled at the agent node level

PR #327, opened 2023-09-01
Add example using llama-index

cbornet:


cbornet:
cc @hemidactylus

examples/applications/llamaindex-cassandra-sink/python/llamaindex_cassandra.py
```diff
@@ -0,0 +1,73 @@
...
+        self.index = None
+
+    def init(self, config: Dict[str, Any]):
+        self.config = config
+        openai.api_key = config['openai-key']
+
+    def start(self):
+        cluster = Cluster(
+            cloud={
+                "secure_connect_bundle": "/app-code-download/python/secure-connect-bundle.zip",
```
eolivelli, line 42: please add a comment to say that this is just an example and you should not store the secure connect bundle zip

eolivelli: please create a agentContext variable to get the path for the application code

eolivelli: it would be great to do as in the other agents and read the secure bundle as base64 encoded

I am telling this because everybody that is trying out LS already has configured the secure bundle that way in the resources file and they already have the secret
like here:

https://github.com/LangStream/langstream/tree/19b8b0b165a0d94e16fb388fd9e971c1405ca6c4/examples/applications/astradb-sink

cbornet: OK. Then I guess we can write a temp file and zip it from the bytes.

cbornet: Actually we can create in-mem file-like objects in python that should be work with the cassandra driver. I'll have a look.

cbornet: done. PTAL again

eolivelli (approved):
Lgtm

Issue #275, opened 2023-08-25
Runtime code download: enable retries


nicoloboschi:
In case the control plane is restarting, the code downloader should retry the download

PR #329, opened 2023-09-01
Add retries on admin http client

nicoloboschi:
Fixes: #275

* Added retries logic in the http client. It affects the code download runtime step and all the CLI calls.
* Unfortunately the java11 http client doesn't offer a retry mechanism so I had to implement one from scratch
By default a http call is retried at max 5 times, with a exponential algo (inital 2 seconds, multiplier 1.5)

### DIFF
--- a/langstream-admin-client/pom.xml
+++ b/langstream-admin-client/pom.xml
@@ -26,8 +26,6 @@
 
   <artifactId>langstream-admin-client</artifactId>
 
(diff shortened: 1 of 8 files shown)


cbornet:
Maybe we could use a more advanced http client than the builtin one ?

eolivelli (approved):

PR #330, opened 2023-09-01
Fix application status while deploying

### LINKED ISSUE #207: Status of deployed app is always ERROR_DEPLOYING
This is due to the pod status check that expects the pod running. In this case, the pod is a job and it's completed

nicoloboschi:
Fixes #207

* Handle job completed as a good state
* Set deleting when the cleanup job starts


eolivelli (approved):
Lgtm

PR #336, opened 2023-09-04
Add CLI version (langstream -V)

### LINKED ISSUE #325: CLI should return a version
Langstream CLI doesn't return a version, but it used to. 
```
➜  langstream git:(main) ✗ ./bin/langstream -V
➜  langstream git:(main) ✗ langstream -V
➜  langstream git:(main) ✗ ./bin/langstream --version
➜  langstream git:(main) ✗ langstream --version
➜  langstream git:(main) ✗ 
```

nicoloboschi:
Fixes #325

* The CLI -V command now reads the jar manifest and get the version (maven version) + git revision
Result is
```
$ langstream -V        
LangStream CLI 0.0.11-SNAPSHOT (f89913a6)
```
* Fixed usage of header instead of description for every command

### DIFF
--- a/langstream-cli/pom.xml
+++ b/langstream-cli/pom.xml
@@ -16,7 +16,8 @@
     limitations under the License.
 
 -->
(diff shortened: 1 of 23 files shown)


eolivelli (changes requested):
usually it is "-v" and not "-V"

nicoloboschi:
> usually it is "-v" and not "-V"

they are both used.

for example, Python uses -V

`-v` often stands for verbose so it might be confusing

also picocli defaults to -V

eolivelli (approved):

PR #337, opened 2023-09-04
[agents] compute-embeddings: Implement batch and async execution

eolivelli:
Sending batches of requests to the API allows you to reduce the number of calls and mitigate the risk to fall into errors due to rate limits.

Summary:
- now the compute-embeddings agent sends API requests in batches and it is fully async
- unfortunately the OpenAI Java client is not async, so actually only Vertex AI and Hugging Face REST are really async
- introduced a new parameter `batch-size` with default value 10 in the compute-embeddings agent
- introduced a new parameter `flush-interval` with default 0 (ms) in the compute-embeddings agent
- the API call is executed on reaching batch-size pending records or flush-interval ms of wait.
- the feature is disabled by default as it could add extra latency to the chatbot pipelines

Interesting things to document:
- you should enable batching (batch-size > 0 and flush-interval > 0) on the pipelines that are not sensitive to latency (like background text processing)
- on the chatbot pipelines it is better to keep the defaults (flush-interval to 0) in order to reduce the latency

```
pipeline:
  - name: "compute-embeddings"
    id: "step1"
    type: "compute-ai-embeddings"
    output: "chunks-topic"
    configuration:
      model: "text-embedding-ada-002" # This needs to match the name of the model deployment, not the base model
      embeddings-field: "value.embeddings_vector"
      text: "{{% value.text }}"
      batch-size: 10
      flush-interval: 1000
```

Backward compatiblity
This PR is 100% backward compatible because we set flush-interval to 0 by default, so the effect is that we always compute only 1 embeddings at a time.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -49,7 +49,7 @@ else
   # Build all artifacts
   ./mvnw install -DskipTests -T 1C -Ddocker.platforms="$(docker_platforms)" -PskipPython
   # Build docker images
-  ./mvnw package -DskipTests -Pdocker -T 1C -Ddocker.platforms="$(docker_platforms)" -PskipPython
+  ./mvnw package -DskipTests -Pdocker  -Ddocker.platforms="$(docker_platforms)" -PskipPython
   docker images | head -n 6
 fi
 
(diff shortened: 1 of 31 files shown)


eolivelli:
I am closing this patch as the approach is totally wrong.

The agent shouldn't expect that something already created batches before it.
With the API in this patch an agent is doing batches that are "smaller" than the batch that has been sent from the upstream agent.
but with async processing it is possible that the upstream agent is sending only 1 record at a time.

So the agent that wants to perform "batching" but accumulate the records locally and process them after some time or after accumulating N records.

eolivelli:
I have reworked the patch, reopening the issue

PR #340, opened 2023-09-04
Add possibility to commit on Kafka consumer out of order

cbornet:


langstream-runtime/langstream-runtime-impl/src/main/python/Pipfile
```diff
@@ -1,6 +1,7 @@
 [packages]
 pyYaml = "*"
 confluent-kafka = "*"
+sortedcontainers = "*"
```
cbornet, line 4: Python doesn't have a native TreeSet equivalent 😞 .
So I have to include that one.
sortedcontainers doesn't pull any transitive deps and there's very little/no chance that the API would break given how we use it. So I think it's fine to use it.

PR #343, opened 2023-09-05
Fix Python Kafka offsets and add tests

cbornet:
* Fix committed offset must be last message offset + 1
* Add on_assign/on_revoke
* Add tests


eolivelli (approved):
lgtm

PR #344, opened 2023-09-05
[gateway][cli] Improvements on startup

eolivelli:
Summary:
- currently the "langstream gateway chat" command is very slow at bootstrap
- one reason is that it gets twice the status of all the pods, this happens because it tries to validate the gateway ids, and it issues a GET /tenant/application that currently returns the full status of the pods

Changes:
- add a new endpoint /tenant/application/describe to get only the description of the application
- do not call the API twice but only once


langstream-cli/src/main/java/ai/langstream/cli/commands/gateway/ChatGatewayCmd.java
```diff
@@ -70,6 +70,7 @@ public class ChatGatewayCmd extends BaseGatewayCmd {
     @Override
     @SneakyThrows
     public void run() {
+        log("Initializing...");
```
nicoloboschi, line 73: remove

nicoloboschi (changes requested):

langstream-cli/src/main/java/ai/langstream/cli/commands/gateway/ChatGatewayCmd.java
eolivelli: done

nicoloboschi (approved):

PR #347, opened 2023-09-05
Upgrade only executors actually impacted by the app changes

nicoloboschi:
* When deploying/updating an app, the CP computes the sha256 checksum for the python and java dir (two different digests) and put them in the s3 object metadata
* Added new endpoint in the control plane /api/applications/{tenant}/{appid}/code/{codeid}/info to get the digest for python and java
```
{
  "digests": {"java": "xxx", "python": "yy"}
}
```
This is not exposed in the CLI since it might not have sense for the user since the archive code id is a non documented concept.

* The runtime deployer job now mounts the cluster config + token (like the runtime pod).
* The runtime deployer job calls the CP to get the digests for the new code archive id and the old one. If one of them changed, the new code archive id is applied to the agents. If not, the old is kept. 

The net effect is that if an agent configuration has not changed and the custom code didn't change, the executors of that agent won't be restarted at all. 

Current applications don't have the digest stored in S3. When updating an existing app with the new runtime code, the application will be restarted in any case if it contains custom code.


langstream-core/src/main/java/ai/langstream/impl/parser/ModelBuilder.java
```diff
@@ -126,18 +149,62 @@ public static ApplicationWithPackageInfo buildApplicationInstance(
...
+            TreeMap<String, String> sortedPyBinaries, Path current, Path rootPath)
+            throws IOException {
 
-        return buildApplicationInstance(applicationContents, instanceContent, secretsContent);
+        if (Files.isRegularFile(current)) {
+            try {
+                final Path relativeCurrent = rootPath.relativize(current);
+                sortedPyBinaries.put(
+                        relativeCurrent.toString(),
+                        Files.readString(current, StandardCharsets.UTF_8));
```
eolivelli, line 192: this doesn't work for binary files, like zip (jar) files
you can keep the byte array here

langstream-core/src/main/java/ai/langstream/impl/parser/ModelBuilder.java
```diff
@@ -126,18 +149,62 @@ public static ApplicationWithPackageInfo buildApplicationInstance(
...
+                            }
+                        }
                     }
                 }
             }
         }
+        final ApplicationWithPackageInfo applicationWithPackageInfo =
+                buildApplicationInstance(applicationContents, instanceContent, secretsContent);
+        if (!sortedJavaBinaries.isEmpty()) {
+            final byte[] data = binariesDigestMapper.writeValueAsBytes(sortedJavaBinaries);
```
eolivelli, line 173: this is not very efficient

it is better to traverse the directory and compute the checksum while reading the files,
you don't need to store the files in memory.
therefore, the algorithm will be more efficient.

you can open a MessageDigest and then pass the contents of the files all in one

eolivelli (changes requested):
I have left one comment about how to compute the digest.
Otherwise LGTM

eolivelli (approved):
LGTM

PR #350, opened 2023-09-05
Add test about Kafka topic retention

eolivelli:
Add a test about Topic retention

```
         module: "module-1"
         id: "pipeline-1"
         topics:
           - name: "input-topic-with-retention"
             config:
               retention.ms: 300000
         pipeline:
           - id: "step1"
             type: "identity"
             input: "input-topic-with-retention"
```


eolivelli:
@mendonk we need a section in the LS docs about how to configure Topic properties like this

PR #354, opened 2023-09-05
Implement fully asynchronous Sinks and graceful pod shutdown (Java runtime)

eolivelli:
Summary:
- rework the internal Sink API to be fully async
- Kafka Topic Sink
- Kafka Connect Sinks
- Cassandra Sink
- Pinecone Sink (not really async)
- now that everything is async we have to wait for all the messages produced by the Source to be processed before shutting down the process (or the test)


eolivelli:
This is still work-in-progress.

Now that everything is fully async some tests are flaky because we don't wait for all the messages to be passed to the sink and completely processed

nicoloboschi (approved):

eolivelli:
I have tested the patch manually under some load.
It is safe to commit it

PR #360, opened 2023-09-06
Add Python logging with same format as Java

### LINKED ISSUE #346: Configure Python log format to be coherent with Java

cbornet:
Fix #346


nicoloboschi (commented):
How the custom processor should call the logger?

cbornet:
The processor can use logging.info or it's own logger. Eg:
```python
import logging
logging.info("hello") # uses the root logger

LOG = logging.getLogger(__name__) # uses a logger with the module name
LOG.info("world")
```

eolivelli (approved):
+1

we need to add a section to the docs here

cbornet:
> we need to add a section to the docs here

Sure. Note that logging is built-in in Python. So there is no surprise here for Python devs.

PR #365, opened 2023-09-06
Replace serializer by one inferred from type if it's ByteArraySerializer

cbornet:
Also changed the headers inferred serializer to mimick Java (the first non-null header type sets the serializer for all subsequent headers).

### DIFF
--- a/langstream-runtime/langstream-runtime-impl/src/main/python/langstream_runtime/kafka_connection.py
+++ b/langstream-runtime/langstream-runtime-impl/src/main/python/langstream_runtime/kafka_connection.py
@@ -39,8 +39,10 @@ from .topic_connector import TopicConsumer, TopicProducer
 LOG = logging.getLogger(__name__)
 STRING_DESERIALIZER = StringDeserializer()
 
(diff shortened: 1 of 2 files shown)


langstream-runtime/langstream-runtime-impl/src/main/python/langstream_runtime/kafka_connection.py
```diff
@@ -341,25 +359,31 @@ def write(self, records: List[Record]):
...
-                    elif isinstance(value, str):
-                        headers.append((key, STRING_SERIALIZER(value)))
-                    elif isinstance(value, bool):
-                        headers.append((key, BOOLEAN_SERIALIZER(value)))
-                    elif isinstance(value, int):
-                        headers.append((key, LONG_SERIALIZER(value)))
-                    elif isinstance(value, float):
-                        headers.append((key, DOUBLE_SERIALIZER(value)))
+                    if value is not None:
+                        if self.header_serializer is None:
```
cbornet, line 363: We could also imagine having a per-header name serializer ?

eolivelli (approved):
LGTM

PR #373, opened 2023-09-07
Cli: use global config file by default

nicoloboschi:
* By default, it uses the config file at `$HOME/.langstream/config`. If it doesn't exist, it gets created.
* User can still pass `--conf` to use a different config file
* conf/cli.yaml is anymore included in the zip/docker since it's not necessary
* default values are now set by the cli directly instead of being retrieved by the file 

This is breaking change since the current config file will be ignored and new one will be used. 

Note that the command ´bin/langstream´ is a bit different from the one that you install. The former injects the --conf parameter by using the env LANGSTREAM_CLI_CONF. Now it defaults to the global one as well.
Since that command is likely to be used during development, I added a info at the cli startup

```
Using global CLI config file. To use the dev config file, set LANGSTREAM_CLI_CONFIG=conf/cli.yaml
```

### DIFF
--- a/bin/langstream
+++ b/bin/langstream
@@ -36,13 +36,8 @@ if [ ! -d langstream-cli/target/cli ]; then
   ./mvnw clean install -DskipTests -pl langstream-cli -am
 fi
 popd > /dev/null
-LANGSTREAM_CLI_CONFIG=${LANGSTREAM_CLI_CONFIG:-""}
-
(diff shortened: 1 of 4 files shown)


eolivelli (changes requested):
I think that conf/cli.yaml is pretty useful.

for local development which values are used ?

eolivelli (approved):
LGTM

PR #376, opened 2023-09-07
Cli: ensure JDK11 compatibility

nicoloboschi:
* Set enforcer to jdk11 for cli and admin client module
* Removed other internal dependencies
* Replaced model object with generic maps


eolivelli (approved):
LGTM

PR #377, opened 2023-09-07
Vector DB Support: download automatically the AstraDB secure bundle

eolivelli:
Summary:
- now that we can use the Astra DevOps API it is easy to download the Secure bundle automatically
- the user must provide the "token" and the "database" in the DataSource configuration
- this works for the query-vector-db, vector-db-sink and the astra-keyspace and cassandra-table assets

configuration.yaml

```
configuration:
  resources:
    - type: "vector-database"
      name: "AstraDatasource"
      configuration:
        service: "astra"
        clientId: "{{{ secrets.astra.clientId }}}"
        secret: "{{{ secrets.astra.secret }}}"
        token: "{{{ secrets.astra.token }}}"
        database: "{{{ secrets.astra.database }}}"
        environment: "{{{ secrets.astra.environment }}}"
```

"environment" is by default PROD and it can be PROD, STAGING, DEV, this is a Datastax only thing, because we use the DEV databases.
 I would keep it because all DataStax developers are using DEV databases and so it is easier for us to change the value to DEV in our secrets.yaml files and we can set PROD by default in the examples

### DIFF
--- a/examples/applications/astradb-sink/README.md
+++ b/examples/applications/astradb-sink/README.md
@@ -23,7 +23,13 @@ This is handled by the 'cassandra-table' assets in the pipeline.yaml file.
 
 ## Configure the pipeline
 
-Update the same file and set username, password and the other parameters.
+Update the secrets.yaml file and set the Astra credentials and the database name:
+- clientId
+- secret
+- token
+- database
+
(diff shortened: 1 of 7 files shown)


NadirJ (approved):

NadirJ:
This is Awesome!

PR #379, opened 2023-09-08
Fix cli using jdk11

nicoloboschi:
* There were some conflicts due to pull requests merged at the same time
* we should set *release* in the compiler
* Fix bin/langstream script


eolivelli (approved):
LGTM

PR #381, opened 2023-09-08
Cli: allow to pass remote file - HTTPS

nicoloboschi:
- Each value (-app, -i, -s) if it starts with https:// will be downloaded and included in the zip. If the request fails the command fails as well
- plain HTTP is not supported

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/applications/AbstractDeployApplicationCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/applications/AbstractDeployApplicationCmd.java
@@ -20,7 +20,9 @@ import ai.langstream.cli.util.LocalFileReferenceResolver;
 import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.IOException;
+import java.net.URI;
 import java.net.URL;
+import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.file.Files;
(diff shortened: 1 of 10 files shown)


langstream-cli/src/main/java/ai/langstream/cli/commands/applications/AbstractDeployApplicationCmd.java
```diff
@@ -333,10 +335,40 @@ private static void addApp(File appDirectory, ZipFile zip, Consumer<String> logg
         logger.accept("app packaged");
     }
 
-    private File checkFileExists(String path) {
+    @SneakyThrows
+    private File checkFileExistsOrDownload(String path) {
         if (path == null) {
             return null;
         }
+        if (path.startsWith("http://") || path.startsWith("https://")) {
```
eolivelli, line 343: "http://" is not more a thing in 2023, we should probably not support it

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/AbstractDeployApplicationCmd.java
```diff
@@ -333,10 +335,40 @@ private static void addApp(File appDirectory, ZipFile zip, Consumer<String> logg
...
+            final HttpRequest request =
+                    HttpRequest.newBuilder()
+                            .uri(URI.create(path))
+                            .version(HttpClient.Version.HTTP_1_1)
+                            .GET()
+                            .build();
+            final Path tempFile = Files.createTempFile("langstram", ".bin");
```
eolivelli, line 350: typo:  langstram

eolivelli (changes requested):

PR #383, opened 2023-09-08
Add support for a .langstreamignore file

cbornet:
Follows the same rules as gitignore.
At the moment, no support for langstreamignore overrides in subfolders.

TIP: set association of .langstreamignore to .gitignore file type in IDE


examples/applications/langchain-source/.langstreamignore
```diff
@@ -0,0 +1,148 @@
...
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# This folder holds the libs built for the target
+# and we need it in the package
+!python/lib/
```
cbornet, line 33: Should we also add `java/lib/` to the exclusions ?

eolivelli (approved):
LGTM

examples/applications/langchain-source/.langstreamignore
cbornet: OK. I'll add a rule for Java

PR #388, opened 2023-09-08
Add serialization of dict and list to json bytes in python Kafka

cbornet:


langstream-runtime/langstream-runtime-impl/src/main/python/langstream_runtime/tests/test_kafka_connection.py
```diff
@@ -322,6 +322,7 @@ def test_serializers():
             ("ByteArraySerializer", True, b"\x01"),
             ("ByteArraySerializer", 42, b"\x00\x00\x00\x00\x00\x00\x00\x2A"),
             ("ByteArraySerializer", 42.0, b"\x40\x45\x00\x00\x00\x00\x00\x00"),
+            ("ByteArraySerializer", {"a": "b", "c": 42.0}, b'{"a": "b", "c": 42.0}'),
```
eolivelli, line 325: is this covering only "dict" ?

eolivelli (approved):
+1

great idea

langstream-runtime/langstream-runtime-impl/src/main/python/langstream_runtime/tests/test_kafka_connection.py
cbornet: Good catch! I should also cover list!

cbornet: done

PR #389, opened 2023-09-08
Enable test mode on API gateway

nicoloboschi:
Changes:
* Added boolean field `allow-test-mode` to the `authentication` gateway field. By default is true.
```
- id: consume-output-auth-github
      type: consume
      topic: output-topic
      authentication:
        provider: github
        configuration:
          clientId: "{{ secrets.github.client-id }}"
        allow-test-mode: true
```
* In the api gateway it's possible to configure the test gateway authentication provider
* The principal values are computed based on the input credentials. The computation is sha256+base64. Better to not "store" test credentials in plain text, but the computed value must be the same for the same token/credential.

There are two new auth implementations:
- *http* that performs a GET to a fixed url passing the header and check the status
- *jwt* that performs JWT validation

They can be used also by the gateway for production authentications. 

* The client needs to pass a `test-credentials` param instead of `credentials` to enter the test mode.

Client connect string example with token:

```
ws://localhost:8091/v1/consume/tenant1/application1/consume?test-credentials=TOKEN
```

* In the api gateway, you have to configure the test auth provider

HTTP
```
application.gateways.auth.test.type=http
application.gateways.auth.test.configuration.base-url=http://xxxx
application.gateways.auth.test.configuration.path-template=/token/{tenant}
application.gateways.auth.test.configuration.headers.myheader=myvalue
application.gateways.auth.test.configuration.accepted-statuses=200,201	
```

JWT
```
application.gateways.auth.test.type=jwt
application.gateways.auth.test.configuration.secretKey=xxx
application.gateways.auth.test.configuration.authClaim=sub
application.gateways.auth.test.configuration.adminRoles=adminuser
```


examples/applications/gateway-authentication/gateways.yaml
```diff
@@ -21,20 +21,20 @@ gateways:
     topic: input-topic
     parameters:
       - sessionId
-    produceOptions:
+    produce-options:
```
eolivelli, line 24: this reads like a breaking change and we have stuff in production, please don't do it

eolivelli (changes requested):

langstream-api-gateway-auth/pom.xml
```diff
@@ -30,5 +30,7 @@
   <modules>
     <module>langstream-google-api-gateway-auth</module>
     <module>langstream-github-api-gateway-auth</module>
+    <module>langstream-jwt-gateway-admin-auth</module>
```
eolivelli, line 33: I would drop "admin" from everywhere

examples/applications/gateway-authentication/gateways.yaml
nicoloboschi: no it's not, it's an alias and they both work

langstream-api-gateway/src/main/java/ai/langstream/apigateway/websocket/AuthenticationInterceptor.java
```diff
@@ -100,30 +133,64 @@ public AuthFailedException(String message) {
...
+            String provider = gatewayRequestContext.adminCredentialsType();
+            if (provider == null && defaultProvider != null) {
+                provider = defaultProvider;
+            }
+            if (provider == null) {
+                throw new AuthFailedException("No admin auth provider specified");
+            }
+
+            final GatewayAdminAuthenticationProvider gatewayAdminAuthenticationProvider =
+                    adminAuthProviders.get(provider);
```
eolivelli, line 165: I think that we shouldn't let the client decide which provider,
there will be only one configured, there is no need to have more

langstream-api-gateway/src/main/java/ai/langstream/apigateway/websocket/handlers/AbstractHandler.java
```diff
@@ -189,16 +189,26 @@ private Gateway extractGateway(
...
-    public GatewayRequestContext validateRequest(
+    public GatewayRequestContextImpl validateRequest(
             Map<String, String> pathVars,
             Map<String, String> queryString,
             Map<String, String> httpHeaders) {
         Map<String, String> options = new HashMap<>();
         Map<String, String> userParameters = new HashMap<>();
 
         final String credentials = queryString.remove("credentials");
+        final String adminCredentials = queryString.remove("admin-credentials");
```
eolivelli, line 200: I don't think there is need to have a different parameter

langstream-api-gateway/src/main/java/ai/langstream/apigateway/websocket/handlers/AbstractHandler.java
```diff
@@ -189,16 +189,26 @@ private Gateway extractGateway(
...
+        final String adminCredentialsType = queryString.remove("admin-credentials-type");
+
+        final Map<String, String> adminCredentialsInputs = new HashMap<>();
 
         for (Map.Entry<String, String> entry : queryString.entrySet()) {
+            if (entry.getKey().startsWith("admin-credentials-input-")) {
```
eolivelli, line 206: I am not sure we need this thing, it seems like a security hole, what's the use case ?

langstream-api/src/main/java/ai/langstream/api/gateway/GatewayAuthenticationProviderRegistry.java
```diff
@@ -42,4 +42,26 @@ public static GatewayAuthenticationProvider loadProvider(
         store.initialize(configuration);
         return store;
     }
+
+    public static GatewayAdminAuthenticationProvider loadAdminProvider(
```
eolivelli, line 46: no need to have a different mechanism

langstream-cli/src/main/java/ai/langstream/cli/commands/gateway/ChatGatewayCmd.java
```diff
@@ -67,6 +67,21 @@ public class ChatGatewayCmd extends BaseGatewayCmd {
             description = "Connect timeout for WebSocket connections in seconds.")
     private long connectTimeoutSeconds = 0;
 
+    @CommandLine.Option(
```
eolivelli, line 70: no need

eolivelli (changes requested):
Nice work but:
- I don't think we need a separate class hierarchy, we can make jwt and http regular authentication providers
- I don't think we need the ability to override the values (if we need it then we have to configure it statically somewhere, not let the client override security related pieces of information)
- I don't think we need to have a different admin-credentials parameter, it is enough to have "credentials"


My idea is that if you enable the system authentication then you gateway accepts also the credentials for the system provider and run the authentication flow.

The question maybe is "how do the gateway know that the client is authenticating with one provider, like Google, or another provider, like the System provider ?"


if that's the problem maybe it is good to have "admin-credentials", but I won't create different class hierarchies for the "System Gateway Authentication" and the regular Gateway Authentication

eolivelli (approved):
LGTM

PR #391, opened 2023-09-10
Allow to return tuples instead of Records in Python agents

cbornet:


eolivelli (approved):
LGTM

eolivelli:
A couple of questions before committing:
is this a backward compatible change ?
should we update the docs ? (in this case please add a label “docs-needed”)

langstream-runtime/langstream-runtime-impl/src/main/python/langstream/util.py
```diff
@@ -29,31 +29,31 @@ def __init__(
         self,
         value,
         key=None,
+        headers: List[Tuple[str, Any]] = None,
```
cbornet, line 32: Small breaking change here for consistency.
Only affects if you were using positional args and not named ones.

PR #393, opened 2023-09-11
[cli] Add a tool to run locally an application on docker

eolivelli:
Goals:
allow users to quickly iterate over the configurations without needing to deploy the application to a kubernetes cluster

Summary:
- new docker image langstream-runtime-tester that allows you to run a single LangStream application on the local enviroment, on a single docker container

How to run the tool:

`bin/langstream docker run -app /path/to/app -i /path/to/instance -s /path/to-secrets --start-broker=true|false --start-s3=true|false`

it is the same command as "apps deploy" but you use "local-run" instead of "deploy".
As the environment is ephemeral you must always pass the instance.yaml and the secrets.yaml files

By default we also start Kafka and Minio, you can control them by using --start-broker and --start-s3

Limitations of the "local mode":
- this is not for production, only for quick iterations
- all the pipelines and agents executors run in the same Java process
- no parallelism
- you always have to pass the "secrets" and the "instance file"
- the system is almost stateless, each time you run you deploy the application from scratch

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.langstream.cli.commands.docker;
+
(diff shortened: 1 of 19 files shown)


langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
```diff
@@ -0,0 +1,151 @@
...
+            String instanceContents,
+            String secretsContents,
+            boolean startBroker,
+            boolean startS3)
+            throws Exception {
+        File tmpInstanceFile = Files.createTempFile("instance", ".yaml").toFile();
+        Files.write(tmpInstanceFile.toPath(), instanceContents.getBytes(StandardCharsets.UTF_8));
+        File tmpSecretsFile = Files.createTempFile("secrets", ".yaml").toFile();
+        Files.write(tmpSecretsFile.toPath(), secretsContents.getBytes(StandardCharsets.UTF_8));
+        String imageName = "langstream/langstream-runtime-tester:latest-dev";
```
nicoloboschi, line 130: we should make this configurable

eolivelli: this should be configurable and it should use "some release", ideally the same version of the CLI

eolivelli:
I am committing this first part of the PR.
I am adding support for the "gateways" in the next one.
and I will also make the version configurable

PR #396, opened 2023-09-12
Ensure existing agents are not affected by deployer config changes

nicoloboschi:
Currently if the deployer gets upgraded/restarted, the new config is applied to ALL the agents/statefulsets. 
That also means that every time the runtime version gets upgraded in the deployer, all the executors restart with the new runtime version.
There are problems with this policy:
1. All the executors get restarted when a system upgrade is performed. This is very bad for the user experience/performance/stability.
2. The executor should continue running that particular version of that the runtime. An agent change could break existing applications. 

For the point 2, in the future it could implemented a flag "auto-update" to get the new runtime image whenever the *application* is upgraded - not the deployer.

Changes:
* Save current configuration (only the relevant and public part) in the agent custom resource status and apply it if it's an update. If it's the statefulset creation (first deploy) it gets the current deployer configuration

### DIFF
--- a/helm/crds/agents.langstream.ai-v1.yml
+++ b/helm/crds/agents.langstream.ai-v1.yml
@@ -45,6 +45,8 @@ spec:
             type: object
           status:
             properties:
+              lastConfigApplied:
+                type: string
               status:
                 properties:
                   status:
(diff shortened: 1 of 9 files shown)


eolivelli (approved):
+1

great catch (and fix)

PR #401, opened 2023-09-12
Change the default provider to OpenAI

cdbartholomew:
The default provider should be OpenAI, not Azure. The average user is more likely to have an OpenAI key, which is quite easy to get, than an Azure key. By defaulting to `openai` there is one less env var to set for in the quick start path.


cdbartholomew:
For our internal development in Azure, we can configure the model name, so we can make it line up with this value.

nicoloboschi (approved):

eolivelli (approved):
LGTM

PR #403, opened 2023-09-13
Use least privilege service account for the runtime pod

nicoloboschi:
At the moment, the runtime pod uses a service account which is able to manipulate configmaps & other resources in the tenant namespace. This is not needed and very dangerous since an application could load a k8s client on behalf of it.

Changes:
* We create another service account for the runtime pods, which is not bound to any role
* In the control plane we accepts service account based on the origin namespace and not the service account
* Added tests on the current OIDC flow
* Added flag to disable k8s authentication. This is because in the API gateway JWT auth there's no need to authenticate k8s service accounts  

 
Follow-up work: 
* Ensure service pods will only have access to download the code - needs to introduce roles on the control plane


eolivelli (approved):
+1

PR #404, opened 2023-09-13
[agentrunner] Split the docker image and clean up the logs

eolivelli:
Summary:

Changes to the docker images:
- split the build of the langstream-runtime image into two steps (langstream-runtime-base and langstream-runtime)
- langstream-runtime-base  contains only the OS, the JDK, Python and some system tools
- with this split rebuilding the images locally and uploading them to minikube is faster as you have to not rebuild 1GB of docker image every time

Changes to logging:
- remove all the logging due to classloading
- add logging every 5 seconds about the number of records in the pipeline and the amount of memory used

### DIFF
--- a/langstream-runtime/langstream-runtime-impl/pom.xml
+++ b/langstream-runtime/langstream-runtime-impl/pom.xml
@@ -27,7 +27,6 @@
   <artifactId>langstream-runtime-impl</artifactId>
 
   <properties>
-    <docker.platforms>linux/amd64</docker.platforms>
     <skipPythonPackage>false</skipPythonPackage>
   </properties>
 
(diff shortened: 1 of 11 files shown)


langstream-runtime/langstream-runtime-impl/pom.xml
```diff
@@ -209,97 +208,6 @@
...
-    <dependency>
-      <groupId>io.netty</groupId>
-      <artifactId>netty-transport</artifactId>
-      <version>${netty.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>io.netty</groupId>
-      <artifactId>netty-transport-native-unix-common</artifactId>
-      <version>${netty.version}</version>
-    </dependency>
```
nicoloboschi, line 296: remove ?

langstream-runtime/langstream-runtime-impl/src/main/docker/Dockerfile
```diff
@@ -58,8 +27,8 @@ ENV NLTK_DATA="/app/nltk_data"
 
 # Install python runtime deps
 RUN cd /app && pipenv requirements --categories="packages full" > /app/requirements.txt \
-    && python3 -m pip install -r /app/requirements.txt \
-    && python3 -m nltk.downloader -d /app/nltk_data punkt averaged_perceptron_tagger
+     && python3 -m pip install -r /app/requirements.txt \
```
nicoloboschi, line 30: since this stuff takes a lot and it's likely to never change during development, can we put it in the base image ?

langstream-runtime/langstream-runtime-impl/pom.xml
eolivelli: this is commented code

langstream-runtime/langstream-runtime-impl/src/main/docker/Dockerfile
eolivelli: but we cannot do it

because of this line, that depends on the langstream-runtime build (the "maven" directory)

ADD maven/Pipfile.lock /app/Pipfile.lock


cc @cbornet

nicoloboschi (approved):

langstream-runtime/langstream-runtime-impl/src/main/docker/Dockerfile
cbornet: I think we should isolate the python stuff in it’s own maven subproject.

PR #406, opened 2023-09-13
Add gRPC processor agent

cbornet:
For preliminary review.
Still need to:
* Put the agent in its own artifact
* Build a NAR
* Resolve TODOs

### DIFF
--- a/langstream-agents/langstream-agent-grpc/src/main/java/ai/langstream/agents/grpc/GrpcAgentProcessor.java
+++ b/langstream-agents/langstream-agent-grpc/src/main/java/ai/langstream/agents/grpc/GrpcAgentProcessor.java
@@ -15,15 +15,14 @@
  */
 package ai.langstream.agents.grpc;
 
-import ai.langstream.agents.grpc.ProcessorGrpc.ProcessorStub;
 import ai.langstream.api.runner.code.AbstractAgentCode;
+import ai.langstream.api.runner.code.AgentContext;
 import ai.langstream.api.runner.code.AgentProcessor;
 import ai.langstream.api.runner.code.RecordSink;
 import ai.langstream.api.runner.code.SimpleRecord;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.protobuf.ByteString;
 import io.grpc.ManagedChannel;
-import io.grpc.ManagedChannelBuilder;
 import io.grpc.stub.StreamObserver;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -32,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.avro.Conversions;
 import org.apache.avro.generic.GenericDatumReader;
@@ -46,12 +46,11 @@ import org.apache.avro.io.EncoderFactory;
 public class GrpcAgentProcessor extends AbstractAgentCode implements AgentProcessor {
 
     protected ManagedChannel channel;
-    private ProcessorStub processorStub;
     private StreamObserver<ProcessorRequest> request;
     private RecordSink sink;
 
     // For each record sent, we increment the recordId
-    private final AtomicInteger recordId = new AtomicInteger(0);
+    private final AtomicLong recordId = new AtomicLong(0);
 
     // For each record sent, we store the record and the sink to which the result should be emitted
     private final Map<Long, RecordAndSink> sourceRecords = new ConcurrentHashMap<>();
@@ -66,8 +65,10 @@ public class GrpcAgentProcessor extends AbstractAgentCode implements AgentProces
     private final Map<Integer, Object> serverSchemas = new ConcurrentHashMap<>();
 
     private final StreamObserver<ProcessorResponse> responseObserver = getResponseObserver();
+    private AgentContext agentContext;
 
-    private record RecordAndSink(ai.langstream.api.runner.code.Record sourceRecord, RecordSink sink) {}
+    private record RecordAndSink(
+            ai.langstream.api.runner.code.Record sourceRecord, RecordSink sink) {}
 
     public GrpcAgentProcessor() {}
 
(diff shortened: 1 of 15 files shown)


langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/GrpcAgentProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+    private final Map<Long, RecordAndSink> sourceRecords = new ConcurrentHashMap<>();
+
+    // For each schema sent, we increment the schemaId
+    private final AtomicInteger schemaId = new AtomicInteger(0);
+
+    // Schemas sent to the server
+    private final Map<Object, Integer> schemaIds = new ConcurrentHashMap<>();
+
+    // Schemas received from the server
+    private final Map<Integer, Object> serverSchemas = new ConcurrentHashMap<>();
```
cbornet, line 77: An untrusted remote server could OOM us by sending a big amount of schemas.
I don't think this is an issue atm since the remote application is owned by the user.
But in the future we may have to put a limit.

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/GrpcAgentProcessor.java
```diff
@@ -0,0 +1,386 @@
...
+        } else if (value.hasJsonValue()) {
+            return value.getJsonValue();
+        } else if (value.hasAvroValue()) {
+            Object serverSchema = serverSchemas.get(value.getSchemaId());
+            if (serverSchema instanceof Schema schema) {
+                return deserializeGenericRecord(schema, value.getAvroValue().toByteArray());
+            } else {
+                // TODO: received unknown schema. error ?
+                // Do we send an error result or do we fail completely so that the agent is
+                // restarted ?
```
cbornet, line 194: What should we do here ?
Ignore ? Send error for the source record ? Fail the producer completely at the next process call ?

eolivelli: fail completely

cbornet: OK.
Also currently I restart the request when the server errors.
But maybe it would be better to also fail completely to have the pod completely restarted.

langstream-agents/langstream-agent-grpc/src/main/java/ai/langstream/agents/grpc/GrpcAgentProcessor.java
```diff
@@ -0,0 +1,366 @@
...
+@Slf4j
+public class GrpcAgentProcessor extends AbstractAgentCode implements AgentProcessor {
+
+    protected ManagedChannel channel;
+    private ProcessorStub processorStub;
+    private StreamObserver<ProcessorRequest> request;
+    private RecordSink sink;
+
+    // For each record sent, we increment the recordId
+    private final AtomicInteger recordId = new AtomicInteger(0);
```
eolivelli, line 54: probably we need AtomicLong here, it is possible to reach 2 billion messages for a production system that is processing data

cbornet: Of course! I use long everywhere for record id (protocol, caches) except here... Good catch !

cbornet: Note that it wouldn't have been a real issue since record ids are transient and forgotten once we get the server response. I believe there would be an OOM before reaching the MAX_VALUE if they are not removed.

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/GrpcAgentProcessor.java
cbornet: done

langstream-agents/langstream-agent-grpc/src/main/java/ai/langstream/agents/grpc/GrpcAgentProcessor.java
cbornet: done

langstream-runtime/langstream-runtime-impl/pom.xml
```diff
@@ -269,6 +269,13 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>langstream-agent-grpc</artifactId>
```
eolivelli, line 274: this is not needed here, please remove it

langstream-api/src/main/java/ai/langstream/api/runner/code/AgentContext.java
```diff
@@ -43,4 +43,9 @@ public interface AgentContext {
     default BadRecordHandler getBadRecordHandler() {
         return DEFAULT_BAD_RECORD_HANDLER;
     }
+
+    default void criticalFailure(Throwable error) {
+        System.err.printf("Critical failure: %s. Shutting down the runtime...", error.getMessage());
```
eolivelli, line 48: for the default implementation we can simply log, this way the tests won't crash the JVM

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
```diff
@@ -1036,6 +1035,15 @@ public BadRecordHandler getBadRecordHandler() {
...
             return topicConnectionProvider;
         }
+
+        @Override
+        public void criticalFailure(Throwable error) {
+            log.error(
+                    "Critical failure: %s. Shutting down the runtime..."
+                            .formatted(error.getMessage()),
+                    error);
+            AgentContext.super.criticalFailure(error);
```
eolivelli, line 1045: We can make the behaviour configurable with a static "Consumer<Throwable>" variable, and in AbstractApplicationRunner (that is used in unit tests) we intercept the call without crashing the JVM

eolivelli (commented):
I have left some final feedback.
Overall LGTM

PR #412, opened 2023-09-14
Split topic/assets deploy in a dedicated job

nicoloboschi:
Changes:

The application deployment now follows this logic
1. The deployer creates a `app-setup` job. This creates the topic and assets. It runs with the non privileged service account.
2. If it fails, the app goes into state ERROR_DEPLOYING
3. If it completes, the deployer creates another job `runtime-deployer`. This creates the agents CRDs.
4. If it fails, the app goes into state ERROR_DEPLOYING
5. If it completes, the app goes into state DEPLOYED

Notes:
* The delete process hasn't been touched
* The agents are now created only if the assets setup went well


langstream-runtime/langstream-runtime-api/src/main/java/ai/langstream/runtime/api/agent/RuntimePodConfiguration.java
```diff
@@ -16,12 +16,10 @@
...
 import ai.langstream.api.model.StreamingCluster;
-import java.util.List;
 import java.util.Map;
 
 public record RuntimePodConfiguration(
         Map<String, Object> input,
         Map<String, Object> output,
         AgentSpec agent,
-        StreamingCluster streamingCluster,
-        List<Map<String, Object>> assets) {}
```
eolivelli, line 27: what happens to existing applications ?

eolivelli (commented):
The change LGTM.

I am worried about existing applications, what's the story ?
1) what happens to an existing application when you restart the pods (or they are restarted by k8s) ?
2) what happens if the upgrade of the operator happens while an application is deploying or being deleted ?

nicoloboschi:
> The change LGTM.
> 
> I am worried about existing applications, what's the story ?
> 
> 1. what happens to an existing application when you restart the pods (or they are restarted by k8s) ?
> 2. what happens if the upgrade of the operator happens while an application is deploying or being deleted ?


The existing app won't be affected. 
When an existing app is updated, the new runtime and job will be created.

The only case that is not fully covered is:
- the application is being deployed with the job
- the deployer gets upgraded before being able to handle the agents
- if the app was declaring assets, they won't be created

To handle this scenario, we could still ensure the assets in the agents runner code. I'm not sure if it's worth it. In that case, a simple apps update will fix the situation.

PR #413, opened 2023-09-14
[kafka] Support writing messages with different schemas

eolivelli:
Summary:
Use a smarter algorithm to cache the Serialiser while producing to a Kafka topic

### DIFF
--- a/langstream-kafka-runtime/src/main/java/ai/langstream/kafka/runner/KafkaProducerWrapper.java
+++ b/langstream-kafka-runtime/src/main/java/ai/langstream/kafka/runner/KafkaProducerWrapper.java
@@ -74,8 +74,13 @@ class KafkaProducerWrapper implements TopicProducer {
     private final AtomicInteger totalIn = new AtomicInteger();
     KafkaProducer<Object, Object> producer;
     Serializer keySerializer;
+    Class cacheKeyForKeySerializer;
     Serializer valueSerializer;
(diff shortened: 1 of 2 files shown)


langstream-kafka-runtime/src/main/java/ai/langstream/kafka/runner/KafkaProducerWrapper.java
```diff
@@ -160,20 +165,30 @@ public CompletableFuture<?> write(Record r) {
...
-                    if (keySerializer == null) {
-                        keySerializer = getSerializer(r.key().getClass(), keySerializers, true);
+                    Class<?> keyClass = r.key().getClass();
+                    if (keySerializer == null
+                            || !(Objects.equals(keyClass, cacheKeyForKeySerializer))) {
+                        keySerializer = getSerializer(keyClass, keySerializers, true);
+                        cacheKeyForKeySerializer = keyClass;
                     }
                     key = keySerializer.serialize(topicName, r.key());
+                    log.info("Chosen serializer for key  {} is {}", keyClass, keySerializer);
```
nicoloboschi, line 175: I'd log this only when the serializer changes

eolivelli: good catch.
it was a left over

PR #418, opened 2023-09-15
Introduce deletion-mode for topics

nicoloboschi:
Changes:

* New option in the topic section `deletion-mode`
Possible values are `none`, `delete`
`none` is the default value.

`none` means to keep the topic when the app is deleted
`delete` means to delete the topic when the app is deleted

Note that `delete` works only if the `creation-mode` is `create-if-not-exists`. This is because in that case we know that we can actually manage that topic inside the langstream app.

Example:
```
topics:
  - name: input-topic
  - name: output-topic
    deletion-mode: delete
```

* Added a new setup-cleanup job right after the runtime deployer cleanup. The latter only deletes the agents. The new job is responsible to delete the topics (and the assets very soon)
* Added the app deletion in the e2e tests to ensure correct behaviour

### DIFF
--- a/langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
+++ b/langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
@@ -110,6 +110,11 @@ public class CassandraAssetsManagerProvider implements AssetManagerProvider {
                 }
             }
         }
+
+        @Override
+        public void deleteAsset() throws Exception {
+            throw new UnsupportedOperationException();
+        }
     }
 
     private static class CassandraKeyspaceAssetManager extends BaseCassandraAssetManager {
@@ -146,6 +151,11 @@ public class CassandraAssetsManagerProvider implements AssetManagerProvider {
                             e.toString());
                 }
             }
+        }
+
+        @Override
+        public void deleteAsset() throws Exception {
+            throw new UnsupportedOperationException();
         }
     }
 
@@ -202,6 +212,11 @@ public class CassandraAssetsManagerProvider implements AssetManagerProvider {
                     throw new IllegalStateException("Error creating keyspace", e);
                 }
             }
+        }
+
+        @Override
+        public void deleteAsset() throws Exception {
+            throw new UnsupportedOperationException();
         }
     }
 
--- a/langstream-core/src/main/java/ai/langstream/impl/common/BasicClusterRuntime.java
+++ b/langstream-core/src/main/java/ai/langstream/impl/common/BasicClusterRuntime.java
@@ -354,6 +354,7 @@ public abstract class BasicClusterRuntime implements ComputeClusterRuntime {
                 new TopicDefinition(
                         name,
                         creationMode,
+                        TopicDefinition.CREATE_MODE_NONE,
                         inputTopicDefinition.isImplicit(),
                         inputTopicDefinition.getPartitions(),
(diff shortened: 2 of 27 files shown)


langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
```diff
@@ -147,6 +152,11 @@ public void deployAsset() throws Exception {
                 }
             }
         }
+
+        @Override
+        public void deleteAsset() throws Exception {
```
eolivelli, line 157: we can add a "default" method implementation here

langstream-core/src/main/java/ai/langstream/impl/common/BasicClusterRuntime.java
```diff
@@ -354,6 +354,7 @@ private Topic buildImplicitTopicForDeadletterQueue(
                 new TopicDefinition(
                         name,
                         creationMode,
+                        TopicDefinition.CREATE_MODE_NONE,
```
eolivelli, line 357: this should be the same as the original topic, the same way we do for "creationMode"

langstream-kafka-runtime/src/main/java/ai/langstream/kafka/runner/KafkaTopicConnectionsRuntime.java
```diff
@@ -265,20 +265,30 @@ public void delete(ExecutionPlan applicationInstance) {
...
-                try {
-                    admin.deleteTopics(List.of(topic.name()), new DeleteTopicsOptions())
-                            .all()
-                            .get();
-                } catch (ExecutionException e) {
-                    if (e.getCause() instanceof UnknownTopicOrPartitionException) {
-                        log.info("Topic {} does not exist", topic.name());
-                    } else {
-                        throw e;
+                if (topic.deleteMode().equals(TopicDefinition.DELETE_MODE_DELETE)) {
```
eolivelli, line 268: topic.deleteMode() may be null for old pipelines, please use Objects.equals()

langstream-kafka-runtime/src/main/java/ai/langstream/kafka/runner/KafkaTopicConnectionsRuntime.java
```diff
@@ -265,20 +265,30 @@ public void delete(ExecutionPlan applicationInstance) {
...
+                } else {
+                    log.info(
+                            "Keeping Kafka topic {} since deletion-mode is {}",
+                            topic.name(),
+                            topic.deleteMode());
                 }
             }
-            default -> log.info("Keeping Kafka topic {}", topic.name());
+            default -> log.info(
+                    "Keeping Kafka topic {} since creation-mode is {}",
```
eolivelli, line 289: this is not clear

eolivelli (changes requested):
I have left some comments

I am not sure that this is enough clear
`deletion-mode: delete`


what about:
`deletion-mode: delete-on-uninstall`

because 'delete' means too less (or too much)

langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
nicoloboschi: this code is not actually called, I'll follow up in the next pr

eolivelli (approved):

PR #421, opened 2023-09-15
Introduce deletion-mode for assets

nicoloboschi:
Same of https://github.com/LangStream/langstream/pull/418 but for assets.

new field `deletion-mode: none|delete`. Default is none

```
assets:
- name: "vsearch-keyspace"
    asset-type: "cassandra-keyspace"
    deletion-mode: delete
    config:
      keyspace: "vsearch"
      datasource: "CassandraDatasource"
      create-statements:
        - "CREATE KEYSPACE vsearch WITH REPLICATION = {'class' : 'SimpleStrategy','replication_factor' : 1};"
```

### DIFF
--- a/langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
+++ b/langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
@@ -60,10 +60,8 @@ public class CassandraAssetsManagerProvider implements AssetManagerProvider {
 
         @Override
         public boolean assetExists() throws Exception {
-            String tableName =
-                    ConfigurationUtils.getString("table-name", null, assetDefinition.getConfig());
-            String keySpace =
-                    ConfigurationUtils.getString("keyspace", null, assetDefinition.getConfig());
+            String tableName = getTableName();
+            String keySpace = getKeySpace();
             log.info("Checking is table {} exists in keyspace {}", tableName, keySpace);
 
             CqlSession session = datasource.getSession();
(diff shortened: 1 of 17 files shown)


langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
```diff
@@ -111,18 +117,31 @@ public void deployAsset() throws Exception {
...
+        }
+
         @Override
-        public void deleteAsset() throws Exception {
-            throw new UnsupportedOperationException();
+        public boolean deleteAssetIfExists() throws Exception {
+            final String keySpace = getKeySpace();
+            List<String> statements =
+                    ConfigurationUtils.getList("delete-statements", assetDefinition.getConfig());
+            if (statements.isEmpty()) {
```
eolivelli, line 131: I don't think this is a good idea, the table may have other dependencies like INDEXes it really depends on how you created it (and we have no control over the create-statements)

if there is no statement then we do nothing, like for create-statements

langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
```diff
@@ -154,8 +182,14 @@ public void deployAsset() throws Exception {
         }
 
         @Override
-        public void deleteAsset() throws Exception {
-            throw new UnsupportedOperationException();
+        public boolean deleteAssetIfExists() throws Exception {
+            List<String> statements =
+                    ConfigurationUtils.getList("delete-statements", assetDefinition.getConfig());
+            if (statements.isEmpty()) {
+                statements = List.of("DROP KEYSPACE IF EXISTS %s;".formatted(getKeyspace()));
```
eolivelli, line 189: also here

eolivelli (changes requested):

langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/cassandra/CassandraAssetsManagerProvider.java
nicoloboschi: it's not very intuitive. 
You set deletion-mode: delete and then you have to provide the keyspace delete statement ?
It's much easier if you just delete it. 

From my testing, the keyspace is being deleted even if not empty and it works well. 

Do you think an INDEX would make the keyspace deletion fail ?

eolivelli (approved):
if the auto delete command works with tables with  indexed (for instance of Vector columns) then I am +1


as an alternative we can fail the deployment here

https://github.com/LangStream/langstream/blob/965c78e7261471743c70dc13cf9e5d9ad316b3e6/langstream-core/src/main/java/ai/langstream/impl/assets/CassandraAssetsProvider.java#L30

in any case we must add the validation in CassandraAssetsProvider

PR #450, opened 2023-09-20
E2E tests: allow to use an external kafka cluster

nicoloboschi:


langstream-e2e-tests/pom.xml
```diff
@@ -52,6 +52,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>langstream-kafka-runtime</artifactId>
```
eolivelli, line 57: please note that this is adding the kafka client dependency and also Kafka Connect with tons of third party deps (like versions of Netty...)

langstream-e2e-tests/src/test/java/ai/langstream/tests/util/kafka/RemoteKafkaProvider.java
```diff
@@ -0,0 +1,89 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.clients.admin.Admin;
+
+@Slf4j
+public class RemoteKafkaProvider implements StreamingClusterProvider {
+
+    private static final String SYS_PROPERTIES_PREFIX = "langstream.tests.remotekafka.props.";
+    private static final Map<String, Object> ADMIN_CONFIG;
+
+    private org.apache.kafka.clients.admin.Admin admin;
```
eolivelli, line 36: are we adding the langstream-kafka-runtime only for this class ?
it is better to add only the kafka-clients jar, and not import all the other dependencies

eolivelli (changes requested):

PR #452, opened 2023-09-20
CLI: Allow updates to the default profile

nicoloboschi:
Currently it's not possible to update the default profile. This is not very handy. There's no strong reason to block the update
Changes:
* Now the default profile can be modified with `update` and `import`. It still cannot be deleted

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/profiles/ImportProfileCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/profiles/ImportProfileCmd.java
@@ -42,7 +42,8 @@ public class ImportProfileCmd extends BaseProfileCmd {
 
     @CommandLine.Option(
             names = {"-u", "--update"},
-            description = "Allow updating the profile if it already exists. Note that the configuration will be overwritten and NOT merged")
+            description =
+                    "Allow updating the profile if it already exists. Note that the configuration will be overwritten and NOT merged")
     private boolean allowUpdate;
 
     @CommandLine.Option(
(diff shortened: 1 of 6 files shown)


langstream-cli/src/main/java/ai/langstream/cli/commands/profiles/ImportProfileCmd.java
```diff
@@ -42,7 +42,8 @@ public class ImportProfileCmd extends BaseProfileCmd {
 
     @CommandLine.Option(
             names = {"-u", "--update"},
-            description = "Allow updating the profile if it already exists")
+            description =
+                    "Allow updating the profile if it already exists. Note that the configuration will be overwritten and NOT merged")
```
eolivelli, line 46: what do you mean with "merged" ?

nicoloboschi: I mean that all fields will be overwritten even if they are blank. 
If the profile has a token set and you import with the same name but with a blank string, it will override the token with the blank string

eolivelli: can't we make it more "merge" like ?
it seems awkward,
it should be easy to implement, as you already have the previous version of the profile

nicoloboschi: this is "import", you can use update for that

PR #457, opened 2023-09-20
[compute-ai-embeddings] Compute Embeddings Agent: execute calls asynchronously and in batches while preserving per-key ordering 

eolivelli:
Summary:

This patch comes with two contributions:
- OrderedAsyncBatchExecutor
- Using OrderedAsyncBatchExecutor in the Compute Embeddings Agent

## OrderedAsyncBatchExecutor

This  patch introduces an utility to implement batch computations, OrderedAsyncBatchExecutor, but with asynchronous completion and strict ordering guarantees.
This utility will be used by all the async agents, especially the ComputeEmbedding agent, the AI Chat completions, the Query  agents and the Vector DB Sink agents.

The idea is that for each record we compute an hash function that maps the record to a bucket. 
Each bucket processes the record in order by passing them in batches to an external processor.

The processor is supposed to be asynchronous and it must acknowledge the completion of the execution for the current batch (by completing a CompletableFuture).

OrderedAsyncBatchExecutor will never send two concurrent batches for the same bucket to the Processor, it always waits for the Processor to finish before submitting the next batch.

This patch also moves TransformFunctionUtil.BatchExecutor to a common module, in order to be used by other components, next to OrderedAsyncBatchExecutor

## Compute Embeddings Agent

The new utility is now used by the "compute-embeddings". The agent already sent batches of texts to the API, but before this patch it waited for each call to complete before sending the next batch.

The default behaviour, with flush-interval = 0, sends only one text at a time, and this is perfect for realtime chat.
When the flush-interval is greater than zero the texts are grouped, taking into consideration the key of the message, then they are sent to the API.

The new parameter "concurrency", with default = 4, defines how many concurrent APIs calls the agent can issue at a time. This works because we use the concurrency parameter to define the number of "buckets" for the underlying  OrderedAsyncBatchExecutor

### DIFF
--- a/langstream-agents/langstream-ai-agents/src/main/java/com/datastax/oss/streaming/ai/ComputeAIEmbeddingsStep.java
+++ b/langstream-agents/langstream-ai-agents/src/main/java/com/datastax/oss/streaming/ai/ComputeAIEmbeddingsStep.java
@@ -15,9 +15,9 @@
  */
 package com.datastax.oss.streaming.ai;
 
+import ai.langstream.api.util.BatchExecutor;
 import com.datastax.oss.streaming.ai.embeddings.EmbeddingsService;
 import com.datastax.oss.streaming.ai.model.JsonRecord;
-import com.datastax.oss.streaming.ai.util.TransformFunctionUtil;
 import com.samskivert.mustache.Mustache;
 import com.samskivert.mustache.Template;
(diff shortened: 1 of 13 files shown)


mendonk:
https://docs.langstream.ai/building-applications/api-reference/agents#compute-ai-embeddings

PR #460, opened 2023-09-20
Allow to return dict instead of Records in Python agents

cbornet:


eolivelli (approved):

eolivelli:
@cbornet can you please rebase ?

PR #468, opened 2023-09-21
Mini langstream

nicoloboschi:
* Added `mini-langstream` script with instructions. By default it installs the latest version of langstream.
It requires:
* Minikube
* kubectl
* helm
* docker
* langstream (cli)

When you run `mini-langstream start`, the following happens:
* Minikube context is setup
* MinIO is created inside a docker container
* Kafka is created inside a docker container
* LangStream is deployed using helm in the minikube context
* PortForwarding is enabled for gateway and control plane
* A dedicated CLI config is created

You can then use the CLI targeting the cluster with:
```
mini-langstream cli apps list
```
* Added to the release procedure
* Added binary installation similar to langstream CLI


bin/mini-langstream/README.md
```diff
@@ -0,0 +1,40 @@
+# mini-langstream
+
+## Get started
+
+### MacOS/Linux/WSL
+
+```
+curl -Ls "https://raw.githubusercontent.com/LangStream/langstream/main/bin/mini-langstream/get-mini-langstream.sh" | bash
```
eolivelli, line 8: Can we make it available in the homebrew formula ?

Personally (maybe it is a matter of taste) I don't like to install stuff this way on my Mac, as there is no way of uninstalling things. With homebrew (and other managers) the whole lifecycle is handled (also the upgrades)

bin/mini-langstream/README.md
```diff
@@ -0,0 +1,40 @@
+# mini-langstream
```
eolivelli, line 1: I am not sure that the README should go into the "bin" directory,
maybe it is better to move the whole directory to the root (not as child of "bin")

eolivelli (changes requested):

bin/mini-langstream/README.md
nicoloboschi: yes this is for Linux users. Brew will come later

bin/mini-langstream/README.md
nicoloboschi: good point

eolivelli (approved):

PR #469, opened 2023-09-21
[webcrawler] Do not fail in case of unknown properties in the JSON file (handle backward compatibility)

eolivelli:
Summary:
- we recently changed the format of the JSON file with the status of the Web Crawler, this broke compatibility and you cannot easily upgrade the application to the newer version
- with this change in case of a broken JSON file (or with an incompatible format) we ignore the file and start from scratch

For docs:
- change the default of max-depth to 50


eolivelli:
FYI @cdbartholomew

PR #475, opened 2023-09-25
Executor pod: set liveness and readiness probe

nicoloboschi:
Currently the executor doesn't have any readiness/liveness probe. 
The main problem is that the application status relies on the statefulset ready status. Therefore an app goes into DEPLOYED even if the executor has not started to consume messages yet.
This problem is more evident with the latest improvements on the classpath. The extraction of NAR agents/streaming runtime can take time in low-resource environments.

As user I'd expect that if an app is DEPLOYED, then I can use gateways and get answers in relevant time. 

Changes:
* This solution introduces readiness/liveness probes that check the metrics endpoint. The http server is setup only after all the executor unpackaging has been done and right before creating the streaming consumer.
* The values are configurable in the deployer configuration


eolivelli (approved):
Lgtm

I would be better to have a dedicated endpoint /healthcheck

But we can change it later

PR #484, opened 2023-09-26
Implement ai-text-completions for VertexAI and HuggingFace inf

nicoloboschi:
* Implemented `ai-text-completions` agent for vertexAI. Only one prompt must be specified. The API is the same and Vertex inference the response depending on the model. Using a text model in the chat agent will cause runtime errors. Using a chat model in the text agent will cause runtime errors.
* Implemented `ai-text-completions` agent for Hugging face. The API is the same as the chat-completions but in this way we support switching between providers while doing tests

### DIFF
--- a/langstream-agents/langstream-ai-agents/src/main/java/ai/langstream/ai/agents/services/impl/VertexAIProvider.java
+++ b/langstream-agents/langstream-ai-agents/src/main/java/ai/langstream/ai/agents/services/impl/VertexAIProvider.java
@@ -275,7 +275,6 @@ public class VertexAIProvider implements ServiceProviderProvider {
                 request.instances.add(instance);
                 appendRequestParameters(additionalConfiguration, request);
 
-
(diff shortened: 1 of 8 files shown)


langstream-agents/langstream-ai-agents/src/main/java/ai/langstream/ai/agents/services/impl/VertexAIProvider.java
```diff
@@ -313,43 +308,82 @@ public CompletableFuture<ChatCompletions> getChatCompletions(
...
                 }
+
+                @Data
+                static class TextInstance {
+                    String prompt;
+                }
             }
 
             @Override
+            @SneakyThrows
```
eolivelli, line 362: Method that returns a CompletableFuture should not throw Exception

eolivelli (changes requested):

PR #488, opened 2023-09-26
Switch Python agents with new gRPC-based ones

cbornet:


eolivelli (approved):
+1

Great work

langstream-e2e-tests/src/test/resources/apps/python-sink/python/lib/confluent_kafka/__init__.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python
```
cbornet, line 1: This also tests the custom user lib import feature !

PR #489, opened 2023-09-27
Allow to configure more than one AI service

nicoloboschi:
Currently if you configure both open-ai and vertex-ai as AI resources, the AI Agent will use the first one loaded at runtime. 

Changes:
* Added support for specifying the service type using `ai-service: x` where x must match a provider id/name

Example:

```yaml
configuration:
  resources:
    - name: open-ai
      type: open-ai-configuration
      configuration:
        url: "http://something"
        access-key: "xxcxcxc"
        provider: "azure"
    - name: my-vertex
      type: vertex-configuration
      configuration:
        url: "http://something"
        token: xx
        project: yy
```

```yaml
module: "module-1"
id: "pipeline-1"
topics:
  - name: "input-topic"
  - name: "output-topic"
pipeline:
  - name: "compute-embeddings"
    id: "step1"
    type: "compute-ai-embeddings"
    input: "input-topic"
    output: "output-topic"
    configuration:
      ai-service: "my-vertex" # refer to the ai-service
      model: "text-embedding-ada-002"
      embeddings-field: "value.embeddings"
      text: "{{% value.name }} {{% value.description }}"
```

### DIFF
--- a/langstream-core/src/main/java/ai/langstream/impl/agents/ai/GenAIToolKitFunctionAgentProvider.java
+++ b/langstream-core/src/main/java/ai/langstream/impl/agents/ai/GenAIToolKitFunctionAgentProvider.java
@@ -280,6 +280,11 @@ public class GenAIToolKitFunctionAgentProvider extends AbstractAgentProvider {
                 }
             };
     private static final Map<String, StepConfigurationInitializer> STEP_TYPES;
+    protected static final String SERVICE_VERTEX = "vertex-configuration";
+    protected static final String SERVICE_HUGGING_FACE = "hugging-face-configuration";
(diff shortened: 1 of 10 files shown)


eolivelli (approved):
LGTM

the original intent was to allow this feature, but it was broken,
thanks for fixing it

cdbartholomew:
Will this change also allow multiple open-ai services? With Azure, you can have different URLs depending on what region you are in, so each needs a different config. In the past I tried to use GPT-3.5 in one region and GPT-4 in another (since it was only available in limited regions) and it didn't work. I had to configure Azure to deploy all models I wanted to use in a single region. The workaround works, but it is a bit cumbersome.

PR #491, opened 2023-09-27
Fix VertexAI parameters

nicoloboschi:
For VertexAI completions the `max-tokens` configuration is required. Currently there's a bug and the vertexAI configuration is not converted correctly from the agent to the request json


eolivelli (approved):
+1

PR #495, opened 2023-09-27
[DO NOT MERGE][breaking] Refactor Python user API

cbornet:
This is a breaking change of the API but it really simplifies usage for the user.
It also brings the possibility to return the process/write results asynchronously.


eolivelli:
Please rebase onto the main branch the other branch has been merged

PR #513, opened 2023-10-02
Generate documentation for resources (AI services, datasource, vector-database)

nicoloboschi:


langstream-core/src/main/java/ai/langstream/impl/resources/AIProvidersResourceProvider.java
```diff
@@ -87,30 +91,165 @@ private void validateVertexConfigurationResource(Resource resource) {
...
+                description =
+                        """
+                        Additional options. DEPRECATED: use "options" in the compute-ai-embeddings agent instead.
+                        """)
+        private Map<String, String> options;
+
+        @ConfigProperty(
+                description =
+                        """
+                        Model name. DEPRECATED: use "model" in the compute-ai-embeddings agent instead.
```
eolivelli, line 251: we can drop this, it has never been used

PR #521, opened 2023-10-03
Update docker-chatbot chart

cbornet:


cbornet:
See changes here: https://github.com/LangStream/langstream/blob/e8188876f2120e8d55f414d0bae9ffe89704bb59/examples/applications/docker-chatbot/README.md#application-flow-chart

PR #527, opened 2023-10-04
Add an example on how to implement websocket clients (javascript)

nicoloboschi:


examples/websockets/openai-completions-chatbot/README.md
```diff
@@ -0,0 +1,3 @@
+# Expose the LangStream application to users via UI + JS Websockets
+
+This is an example on how to implement a chatbot UI using LangStream api gateway endpoints to connect to the LangStream application.
```
eolivelli, line 3: we can suggest users to use the built-in Python http server to run serve code

https://github.com/LangStream/langstream/blob/e8ed1c4e54b473e1f300d0aa2b1a6a970f4094ea/examples/applications/gateway-authentication/README.md#L61

nicoloboschi: done thanks

PR #530, opened 2023-10-04
Cli: include .bat file in the tarball

nicoloboschi:
We can include the generated .bat file in the zip. 
I tested it on windows 11 with java 11 and it works like a charm.

The only problem is that windows users will have to download the tarball, extract it and run it. There's no global command


eolivelli:
also the need to have JDK11 installed

eolivelli (approved):
Great !

PR #533, opened 2023-10-05
Add dry-run apps deployment

### LINKED ISSUE #532: Cli: add command to resolve placeholders locally
It could be very useful to have a CLI command that resolves all the placeholder, for debugging purposes. It would be similar to [Helm template](https://helm.sh/docs/helm/helm_template/).

The output would be sent to stdout.

nicoloboschi:
Fixes #532

Changes:
* New option in the deploy `langstream apps deploy --dry-run my-app -app ...`. This sends a request to the control plane.
* The control plane handles the new option and generates the application description with the resolved application. The app is not deployed
* You can customize the output with `-o yaml/json`. By default is `yaml`
* The same option is not recognized for the update command since we don't want to expose already stored secrets to the user (that might not have access to them)

### DIFF
--- a/langstream-admin-client/src/main/java/ai/langstream/admin/client/AdminClient.java
+++ b/langstream-admin-client/src/main/java/ai/langstream/admin/client/AdminClient.java
@@ -182,15 +182,20 @@ public class AdminClient implements AutoCloseable {
 
     private class ApplicationsImpl implements Applications {
         @Override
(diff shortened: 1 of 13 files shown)


eolivelli (dismissed):
Great work

Can you please add this option also in the "docker run" mode ?
People that are used to use "docker run" don't have a control plane running

eolivelli:
LGTM as far as we add support for "docker run" users.

PR #536, opened 2023-10-05
[agents] New agent 'dispatch' for message routing

### LINKED ISSUE #454: Add agent that can write to one of many possible topics
Can we create an agent that needs to run some logic or call an external API and based on the results, it decides which topic to write to? For example an agent that need to call a REST API for a multiclass classification model, and based on model prediction, the agent decides which topic it will write to from a list of available topics.

eolivelli:
Fixes #454

Summary:
- this patch adds a new processor 'dispatch' that is able to dispatch the record to different topics depending on a condition

This is an example:

```
topics:
  - name: "input-topic"
  - name: "topic1"
  - name: "topic2"
  - name: "default-topic"
pipeline:
  - name: "Dispatch"
    type: "dispatch"
    input: input-topic
    output: default-topic
    id: step1
    configuration:
      routes:
         - when: properties.language == "en"
           destination: topic1
         - when: properties.language == "fr"
           destination: topic2
           action: dispatch
         - when: properties.language == "none"
           action: drop
```

Important notes:
- the routes are evaluated in order, as soon as one condition is verified the message is sent to the destination and the processing stops
- if no route matches then the record is sent to the "output", that can be a topic or another agent, as usual
- for each route you can define an action: dispatch or drop, the default is "dispatch" that means to send the message, with "drop" the message is silently discarded (not emitted to the downstream agent)

### DIFF
--- a/langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
+++ b/langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
@@ -100,10 +100,14 @@ public class DispatchAgent extends AbstractAgentCode implements AgentProcessor {
                 boolean test = r.predicate.test(context);
                 if (test) {
                     if (r.destination.isEmpty()) {
-                        log.info("Discarding record {} - empty destination", record);
+                        if (log.isDebugEnabled()) {
+                            log.debug("Discarding record {} - empty destination", record);
+                        }
                         recordSink.emit(new SourceRecordAndResult(record, List.of(), null));
                     } else {
-                        log.info("Dispatching record {} to topic {}", record, r.destination);
+                        if (log.isDebugEnabled()) {
+                            log.debug("Dispatching record {} to topic {}", record, r.destination);
+                        }
                         TopicProducer topicProducer = producers.get(r.destination);
                         topicProducer
                                 .write(record)
@@ -130,7 +134,9 @@ public class DispatchAgent extends AbstractAgentCode implements AgentProcessor {
                 }
             }
 
-            log.info("Sending record {} to the default destination", record);
+            if (log.isDebugEnabled()) {
+                log.debug("Sending record {} to the default destination", record);
+            }
             recordSink.emit(new SourceRecordAndResult(record, List.of(record), null));
         } catch (Throwable error) {
             log.error("Error processing record: {}", record, error);
--- a/langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
+++ b/langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
@@ -63,6 +63,10 @@ public class DispatchAgent extends AbstractAgentCode implements AgentProcessor {
                     }
                     if (drop) {
                         log.info("Condition: \"{}\", action = drop", when);
+                        if (!destination.isEmpty()) {
+                            throw new IllegalStateException(
+                                    "drop action cannot have a destination");
+                        }
                     } else {
                         log.info(
                                 "Condition: \"{}\", action = dispatch, destination: {}",
--- a/langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
+++ b/langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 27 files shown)


langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/FlowControlAgentsCodeProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+package ai.langstream.agents.flow;
+
+import ai.langstream.api.runner.code.AgentCodeProvider;
+import ai.langstream.api.runner.code.AgentProcessor;
+import java.util.Map;
+import java.util.function.Supplier;
+
+public class FlowControlAgentsCodeProvider implements AgentCodeProvider {
```
nicoloboschi, line 23: this should DispatcherAgentsCodeProvider ?

langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void init(Map<String, Object> configuration) {
+        List<Map<String, Object>> routes =
+                (List<Map<String, Object>>) configuration.getOrDefault("routes", List.of());
+        routes.forEach(
+                r -> {
+                    String when = ConfigurationUtils.getString("when", "", r);
+                    String destination = ConfigurationUtils.getString("destination", "", r);
```
nicoloboschi, line 51: what about using another param "drop" ? 
It could be error-prone this configuration

so then either "destination" or "drop" are required and you can fail the deploy

nicoloboschi (changes requested):
LGTM, just not very sure about the empty destination

langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/FlowControlAgentsCodeProvider.java
eolivelli: the module is "FlowControlAgents", like "TextProcessingAgentsProvider", we will probably add more soon

langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
eolivelli: having two different options may make difficult to use a variable for this value, think about a configurable application with global variables.

eolivelli:
@nicoloboschi I added support for "action: dispatch|drop" PTAL

langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/FlowControlAgentsProvider.java
```diff
@@ -0,0 +1,137 @@
...
+                description =
+                        """
+                        Condition to activate the route.
+                                """)
+        String when;
+
+        @ConfigProperty(
+                description =
+                        """
+                        Destination of the message. If the destination is empty the message is discarded
```
nicoloboschi, line 125: please update here as well

langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/FlowControlAgentsProvider.java
```diff
@@ -0,0 +1,137 @@
...
+            case DISPATCH -> DispatchConfig.class;
+            default -> throw new IllegalArgumentException("Unsupported agent type: " + type);
+        };
+    }
+
+    @AgentConfig(
+            name = "Text extractor",
+            description =
+                    """
+            Extracts text content from different document formats like PDF, JSON, XML, ODF, HTML and many others.
```
nicoloboschi, line 101: Can you fix it ?

langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
```diff
@@ -0,0 +1,174 @@
...
```
nicoloboschi, line 51: can we check for blank text and raise error if action == dispatch

langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/FlowControlAgentsProvider.java
eolivelli: sure

langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/FlowControlAgentsProvider.java
eolivelli: sure

langstream-agents/langstream-agents-flow-control/src/main/java/ai/langstream/agents/flow/DispatchAgent.java
eolivelli: sure, also on the planner side

PR #537, opened 2023-10-05
New http-request agent

nicoloboschi:
* Added new `http-request` agent
* It accepts
  * url: http(s) url to call without query-string
  * query-string (map): accepts string as template. encoding is done automatically
  * method: "get" by default
  * headers (map):  accepts string as template e.g. X-Custom-Header: {{{ value.id }}}
  * body: body to send, must be a string:  accepts string as template e.g. X-Custom-Header: {{{ value.id }}}
  * output-field: field where to output in the record. in case the response return content-type: application/json it will be deserialized as map. otherwise will be threated as string 
  * allow-redirects: set to false to never redirect
  * handle-cookies: set to false to not manage cookies during redirects

Example with authentication
```
                                  - name: "http-request"
                                    type: "http-request"
                                    input: input-topic
                                    output: output-topic
                                    id: step1
                                    configuration:
                                        output-field: value.api
                                        url: "https://somewhere/api/test/{{{ value.id }}}"
                                        method: POST
                                        body: '{"id": "{{{ value.id }}}"}'
                                        headers:
                                          Content-Type: application/json
                                          Authorization: Bearer {{{ secrets.s1.token }}}
```

### DIFF
--- a/langstream-agents/langstream-agent-http-request/src/main/java/ai/langstream/agents/queryhttp/HttpRequestAgent.java
+++ b/langstream-agents/langstream-agent-http-request/src/main/java/ai/langstream/agents/queryhttp/HttpRequestAgent.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.langstream.agents.queryhttp;
+
+import ai.langstream.ai.agents.commons.JsonRecord;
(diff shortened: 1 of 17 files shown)


eolivelli (changes requested):
I would make the query string parameters configurable, otherwise you have to provide a way to perform escaping.

Also I would make the url a constant in the first iteration, this way we prevent malicious usage, like calling URLs forged by a chatbot user

langstream-agents/langstream-agent-http-request/src/main/java/ai/langstream/agents/queryhttp/HttpRequestAgent.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    public void init(Map<String, Object> configuration) {
+        final String url =
+                ConfigurationUtils.requiredNonEmptyField(
+                        configuration, "url", () -> "http-request agent");
+        this.outputFieldName =
+                ConfigurationUtils.requiredNonEmptyField(
+                        configuration, "output-field", () -> "http-request agent");
+        this.method = ConfigurationUtils.getString("method", "GET", configuration);
+        final String body = ConfigurationUtils.getString("body", null, configuration);
```
eolivelli, line 72: pleae make the user agent configurable, also the handling of redirections and cookies, like we do in the webcrawler

eolivelli (approved):
LGTM

cbornet:
It can be interesting to look at what features LangStream loaders provide when we create these Sources.

PR #546, opened 2023-10-06
Add Azure blob storage source

nicoloboschi:
Changes:
* New source `azure-blob-storage-source` that reads azure blobs
* The auth properties are the same as the code storage config
* The properties are idle-time and file-extensions like the s3-source


examples/applications/http-request-processor/README.md
```diff
@@ -0,0 +1,19 @@
+# Http Request processor
```
eolivelli, line 1: This seems unrelated

langstream-agents/langstream-agent-azure-blob-storage-source/src/main/java/ai/langstream/agents/azureblobstorage/source/AzureBlobStorageSource.java
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        @Override
+        public String origin() {
+            return null;
+        }
+
+        @Override
+        public Long timestamp() {
+            return System.currentTimeMillis();
```
eolivelli, line 229: This should be set at instantiation time, otherwise different calls with return different results

eolivelli (approved):
The azure part lgtm
But you still have some code from the http request processor

examples/applications/http-request-processor/README.md
nicoloboschi: yes it's based on the http-request branch, Ill put this in draft until we get the other one merged

langstream-agents/langstream-agent-azure-blob-storage-source/src/main/java/ai/langstream/agents/azureblobstorage/source/AzureBlobStorageSource.java
nicoloboschi: interesting, I copied it from S3 source

PR #551, opened 2023-10-09
New UI for apps monitoring and testing

nicoloboschi:
## Standalone UI for apps
New command `langstream apps ui <appid>` that:
* Starts a webserver locally - using Undertow
* Serves a static html file that shows the logs and let user to connect to the gateways

The gateway connection is proxied by the webserver in order to being able to handle http -> https and pass around browser default policies.
Same for logs.

## Local run UI
The same server is now deployed when running langstream docker run.
There's an option to skip it `--start-ui false`

The output shown in the logs is the command output.
The output is redirect to a temp file and both the UI and the terminal tails that file. 

Not implemented (yet):
- Gateway authentication
- "chat" type gateways


eolivelli (approved):
LGTM

PR #553, opened 2023-10-09
Fix WebCrawler handle-cookies option

### LINKED ISSUE #540: WebCrawler: handle-cookies is not used
handle-cookies: false is not respected, cookies are always kept

nicoloboschi:
Fixes #540


eolivelli (approved):
LGTM

PR #557, opened 2023-10-09
Improve Langchain document example

cbornet:
Loader and splitter are now separated into a Source and a Processor


eolivelli (approved):
Great

PR #568, opened 2023-10-11
Adding example of web crawling chatbot using RAG and chat memory

cdbartholomew:


eolivelli (approved):
Great stuff.

You may want to follow up with:
- adding default values for globals in the configuration.yaml
- switch to native mustache (remove the percent symbols from the templates, they are no more needed)

cdbartholomew:
Yes, I am working on fixing up the variables to use the new format. I haven't been able to get this to work with 0.1.0 yet, so this example is not very useful yet unless you are using an older version of LangStream.

PR #573, opened 2023-10-12
Update variable format for chatbot-rag-memory example

cdbartholomew:
Updates to chatbot-rag-memory example:
* Small tweaks to support the 0.1.0 release
* Switch globals to new variable format
* Switch mustache templating to new format
* Add defaults for globals in configuration.yaml


eolivelli (approved):
Looks good

PR #574, opened 2023-10-12
Add opensearch support

nicoloboschi:
Changes:
* New datasource "opensearch" that support OpenSearch v2 and AWS OpenSearch Serverless
* New asset "opensearch-index" that creates an index on opensearch. mappings and settings are fully configurable
* "query-vector-db" can now query on a opensearch datasource. The query object is fully configurable-
* "vector-db-sink" can now writes to a opensearch datasource. Only bulk is supported with batch-size and flush-interval. All the bulk options are supported (e.g. refresh, timeout..)
* Added example in the examples directory

FULL Rag example
```
topics:
  - name: "insert-topic"
assets:
  - name: "os-index"
    asset-type: "opensearch-index"
    config:
        index-name: "my-index-1"
        datasource: "OSDatasource"
        settings: |
            {
                "index": {
                      "knn": true,
                      "knn.algo_param.ef_search": 100
                }
            }
        mappings: |
            {
                "properties": {
                      "content": {
                            "type": "text"
                      },
                      "embeddings": {
                            "type": "knn_vector",
                            "dimension": 3
                      }
                }
            }
pipeline:
  - id: write
    name: "Write"
    type: "vector-db-sink"
    input: "insert-topic"
    configuration:
      datasource: "OSDatasource"
      index-name: "my-index-1"
      bulk-parameters:
        refresh: "true"
      id: "key"
      fields:
        - name: "content"
          expression: "value.content"
        - name: "embeddings"
          expression: "value.embeddings"
---
topics:
  - name: "input-topic"
  - name: "result-topic"
pipeline:
  - id: read
    name: "read"
    type: "query-vector-db"
    input: "input-topic"
    output: "result-topic"
    configuration:
      datasource: "OSDatasource"
      query: |
        {
          "query": {
            "knn": {
              "embeddings": {
                "vector": ?,
                "k": 1
              }
            }
          }
        }
      fields:
        - "value.embeddings"
      output-field: "value.query-result"
```


langstream-agents/langstream-vector-agents/src/main/java/ai/langstream/agents/vector/opensearch/OpenSearchAssetsManagerProvider.java
```diff
@@ -0,0 +1,178 @@
...
+            String indexName = getIndexName();
+            return datasource
+                    .getClient()
+                    .indices()
+                    .exists(new ExistsRequest.Builder().index(indexName).build())
+                    .value();
+        }
+
+        private String getIndexName() {
+            String tableName =
```
eolivelli, line 76: Nit: indexName

eolivelli (approved):
Over LGTM 
Please add a example application, like the one for solr

PR #580, opened 2023-10-13
Fix pulsar runtime depency conflicts

cbornet:
NB: this makes Pulsar not working anymore with Avro manipulating agents. But without this Pulsar doesn't work at all.
We can fix the problem with Avro later.


cbornet:
Other NB: trying to use non-shaded versions for both the admin and the client doesn't work bc of some weird issue in the class loading.

eolivelli (approved):
+1

PR #583, opened 2023-10-13
Add vector database sink config validation and documentation

nicoloboschi:
Vector databases configuration were not validated and documented because it's per datasource' service
Changes: 
* Added doc for each implementation

### DIFF
--- a/langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/vectors/SolrVectorDatabaseSinkConfig.java
+++ b/langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/vectors/SolrVectorDatabaseSinkConfig.java
@@ -1,0 +1,52 @@
+package ai.langstream.runtime.impl.k8s.agents.vectors;
+
+import ai.langstream.api.doc.AgentConfig;
+import ai.langstream.api.doc.ConfigProperty;
+import ai.langstream.runtime.impl.k8s.agents.QueryVectorDBAgentProvider;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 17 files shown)


langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/vectors/SolrVectorDatabaseSinkConfig.java
```diff
@@ -0,0 +1,66 @@
...
+import ai.langstream.api.doc.AgentConfig;
+import ai.langstream.api.doc.ConfigProperty;
+import ai.langstream.runtime.impl.k8s.agents.QueryVectorDBAgentProvider;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import lombok.Data;
+
+@Data
+@AgentConfig(
+        name = "Solr",
```
eolivelli, line 27: Apache Solr

langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/vectors/SolrVectorDatabaseSinkConfig.java
```diff
@@ -0,0 +1,66 @@
...
+        description =
+                """
+    Writes data to Solr service.
+    The collection-name is configured at datasource level.
+""")
+public class SolrVectorDatabaseSinkConfig
```
eolivelli, line 33: The maim class is not VectorDatabaseSink.
I would change here and everywhere to VectorDatabaseWriter (or something like that)

eolivelli (changes requested):
Very good.
I left a comment to the name of the new classes, they should match the existing base interface for Writers

PR #588, opened 2023-10-15
Support gateway start from partitioned topics offsets

cbornet:


eolivelli (approved):
LGTM

can we follow up with some unit tests that represent a client that is reconnecting and verify that we are not receiving duplicates and we don't lose massages?

cbornet:
Yes. There are some tests for Kafka that we can also run on Pulsar. The problem is that atm, there are class loader issues when running them for Pulsar. I'm working on that.

PR #590, opened 2023-10-16
Add ability to provision persistent disks to agents

eolivelli:


langstream-core/src/test/java/ai/langstream/model/parser/ResourcesSpecsTest.java
```diff
@@ -140,4 +143,92 @@ private static String buildInstanceYaml() {
...
+                            type: "noop"
+                            resources:
+                               disk:
+                                enabled: true
+                                type: "ssd"
+                          - name: "step5"
+                            type: "noop"
+                            resources:
+                               disk:
+                                enabled: true
```
nicoloboschi, line 186: enabled ? it seems redundant

eolivelli: because maybe you want to temporary disable it without removing the configuration.

it is hard to distinguish "not enabled" from nulls that we have in all the code paths that deal with disks, like merging configurations with defaults or handling backward/forward compatibility

nicoloboschi:
I continued the work: 
Changes:
* We compute the disks syntax and send it to the agent crd
* In the operator you can configure the mapping between a type and storage class. Also you can define default storage class name and default disk size. Note that the storage class must already exists.

PR #596, opened 2023-10-16
CLI docker run: create a dedicated profile  

nicoloboschi:
Changes: 
* use "default" as mock tenant, this is likely to match the default profile when then running gateway commands
* creates/updates a profile named "local-docker-run" that points to the right tenant/urls
* Render markdown is checked by default

### DIFF
--- a/conf/cli.yaml
+++ b/conf/cli.yaml
@@ -14,6 +14,16 @@
 # limitations under the License.
 #
 
-webServiceUrl: http://localhost:8090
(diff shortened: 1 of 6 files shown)


eolivelli (approved):

nicoloboschi:
after some thoughts, I'm not sure this is enough. If we set a mock tenant, then to connect using the gateway is not easy. We should also add a dedicated profile in this pr

PR #602, opened 2023-10-17
Add e2e tests for gateway with Pulsar

cbornet:


langstream-api-gateway/pom.xml
```diff
@@ -112,6 +112,11 @@
 			<artifactId>kafka</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.testcontainers</groupId>
+			<artifactId>pulsar</artifactId>
```
eolivelli, line 117: please note that this dependency imports some version of the Pulsar client, is it polluting the classpath ?

eolivelli (approved):
LGTM

nicoloboschi (approved):

langstream-api-gateway/pom.xml
cbornet: It seems to be OK. Testcontainers uses the same shaded jar that the runtime uses.

PR #539, opened 2023-10-05
Add support for PDF, doc, ppt, and other file types in Python/LangChain

cbornet:

PR #603, opened 2023-10-17
Fix runtime image size

nicoloboschi:
Partial revert of: https://github.com/LangStream/langstream/pull/539
basically the ecosystem of "unstructred" plugins is ~6G so we cannot really use all of them


eolivelli (approved):
+1

PR #604, opened 2023-10-17
Cli: fix docker run on linux

nicoloboschi:
Changes: 
* Set generated files readable for every user to avoid permissions error. The files are copied to $HOME/.langstream/tmp directory which is protected to other unix machine users but still readable from the container
* Improve the installation script error message

After these changes, I successfully setup an amazon linux machine to run langstream (I'll add it to the doc):
```
sudo yum update
sudo yum install -y docker jq unzip  java-11-amazon-corretto-headless
sudo usermod -a -G docker ec2-user
id ec2-user
newgrp docker
sudo systemctl enable docker.service
sudo systemctl start docker.service
curl -Ls "https://raw.githubusercontent.com/LangStream/langstream/main/bin/get-cli.sh" | bash
source ~/.bashrc
export OPEN_AI_ACCESS_KEY=your-key-here
langstream docker run test -app https://github.com/LangStream/langstream/blob/main/examples/applications/openai-completions -s https://github.com/LangStream/langstream/blob/main/examples/secrets/secrets.yaml
```

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/BaseCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/BaseCmd.java
@@ -441,7 +441,6 @@ public abstract class BaseCmd implements Runnable {
                             + response.body());
         }
         Files.write(tempFile, response.body());
-        tempFile.toFile().setReadable(true, false);
         final long time = (System.currentTimeMillis() - start) / 1000;
         logger.accept(String.format("downloaded remote file %s (%d s)", path, time));
         return tempFile.toFile();
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/BaseCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/BaseCmd.java
@@ -441,6 +441,7 @@ public abstract class BaseCmd implements Runnable {
                             + response.body());
         }
         Files.write(tempFile, response.body());
+        tempFile.toFile().setReadable(true, false);
         final long time = (System.currentTimeMillis() - start) / 1000;
         logger.accept(String.format("downloaded remote file %s (%d s)", path, time));
         return tempFile.toFile();
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
@@ -27,9 +27,11 @@ import ai.langstream.cli.commands.applications.MermaidAppDiagramGenerator;
 import ai.langstream.cli.commands.applications.UIAppCmd;
 import ai.langstream.cli.util.LocalFileReferenceResolver;
 import java.io.File;
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.attribute.FileAttribute;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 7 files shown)


langstream-cli/src/main/java/ai/langstream/cli/commands/BaseCmd.java
```diff
@@ -441,6 +441,7 @@ public static File downloadHttpsFile(
                             + response.body());
         }
         Files.write(tempFile, response.body());
+        tempFile.toFile().setReadable(true, false);
```
eolivelli, line 444: is this file containing secrets ? it should not be readable by anyone, but only from the user.

I also suggest to use the NIO UnixFileAttributes to perform this operation, and not the legacy File API

langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
```diff
@@ -281,12 +282,10 @@ private void executeOnDocker(
...
             boolean dryRun)
             throws Exception {
-        File tmpInstanceFile = Files.createTempFile("instance", ".yaml").toFile();
-        Files.write(tmpInstanceFile.toPath(), instanceContents.getBytes(StandardCharsets.UTF_8));
+        File tmpInstanceFile = createReadableTempFile("instance", instanceContents);
         File tmpSecretsFile = null;
         if (secretsContents != null) {
-            tmpSecretsFile = Files.createTempFile("secrets", ".yaml").toFile();
-            Files.write(tmpSecretsFile.toPath(), secretsContents.getBytes(StandardCharsets.UTF_8));
+            tmpSecretsFile = createReadableTempFile("secrets", secretsContents);
```
eolivelli, line 288: if this file is containing secrets then it should be readable only by the current user

eolivelli (changes requested):

nicoloboschi:
This might be a better solution https://stackoverflow.com/a/57776923/7309769

langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
```diff
@@ -241,6 +249,17 @@ public void run() {
 
         updateLocalDockerRunProfile(tenant);
 
+        Runtime.getRuntime()
+                .addShutdownHook(
+                        new Thread(
+                                () -> {
+                                    log("Cleaning environment");
```
eolivelli, line 256: can we log this only if there are "temporary files"

langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
```diff
@@ -45,6 +51,8 @@ public class LocalRunApplicationCmd extends BaseDockerCmd {
 
     protected static final String LOCAL_DOCKER_RUN_PROFILE = "local-docker-run";
 
+    private static final Set<Path> temporaryFiles = new HashSet<>();
```
eolivelli, line 54: use a Concurrent set, this structure is accessed from multiple thread

eolivelli (approved):

langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
nicoloboschi: done

langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
nicoloboschi: done

PR #608, opened 2023-10-18
Add support for topic schemas in Pulsar

cbornet:


langstream-pulsar-runtime/src/main/java/ai/langstream/pulsar/runner/PulsarTopicConnectionsRuntimeProvider.java
```diff
@@ -524,19 +532,51 @@ private class PulsarTopicProducer<K> implements TopicProducer {
...
             @SneakyThrows
             public void start() {
-                String topic = (String) configuration.remove("topic");
-                schema = (Schema<K>) configuration.remove("schema");
-                if (schema == null) {
-                    schema = (Schema) Schema.STRING;
+                if (configuration.containsKey("valueSchema")) {
+                    SchemaDefinition valueSchemaDefinition =
+                            mapper.convertValue(
+                                    configuration.remove("valueSchema"), SchemaDefinition.class);
```
eolivelli, line 567: is is safe to "remove" ?

eolivelli (approved):
+1

I have left one comment

langstream-pulsar-runtime/src/main/java/ai/langstream/pulsar/runner/PulsarTopicConnectionsRuntimeProvider.java
cbornet: we work on a copy.
It's also because we call producer.load(configuration) later.

eolivelli: thanks for your clarification

PR #615, opened 2023-10-18
[docker run] Exit the container in case of errors

eolivelli:
Summary:
- ensure that we exit the container in case of any errors: during setup and in case of any agent failure
- reduce some logs
- ensure that logback configuration is applied


nicoloboschi (commented):
LGTM

PR #617, opened 2023-10-18
Add Rag with AWS example

nicoloboschi:


examples/applications/rag-aws/chatbot.yaml
```diff
@@ -0,0 +1,95 @@
...
+                "vector": ?,
+                "k": 10
+              }
+            }
+          }
+        }
+      fields:
+        - "value.embeddings_vector"
+      output-field: "value.related_documents"
+  - name: "ai-chat-completions"
```
eolivelli, line 58: Is it possible to add re-rank here? 
In order to do it we have to also get the embeddings for each retrieved document.

re-rank is very important because it allows you to select more docs from the database and then refine the search

nicoloboschi: done, thanks

PR #618, opened 2023-10-18
Add support for DLT with Pulsar

cbornet:


langstream-api/src/main/java/ai/langstream/api/runtime/ExecutionPlan.java
```diff
@@ -145,7 +145,7 @@ public Topic getTopicByName(String name) {
 
     public TopicDefinition getTopicDefinitionByName(String name) {
         return topics.entrySet().stream()
-                .filter(e -> e.getKey().getName().equals(name))
+                .filter(e -> e.getValue().topicName().equals(name))
```
cbornet, line 148: We lookup by topic name instead of  TopicDefinition name, Is it OK?

langstream-core/src/main/java/ai/langstream/impl/common/BasicClusterRuntime.java
```diff
@@ -342,7 +342,7 @@ private Topic buildImplicitTopicForDeadletterQueue(
             StreamingClusterRuntime streamingClusterRuntime,
             ExecutionPlan physicalApplicationInstance) {
         // connecting two agents requires an intermediate topic
-        String name = connection.topicName() + "-deadletter";
+        String name = inputTopicDefinition.getName() + "-deadletter";
```
cbornet, line 345: We use the TopicDefinition name instead of the topic name. Is it OK?

langstream-api/src/main/java/ai/langstream/api/runtime/ExecutionPlan.java
eolivelli: if all the tests are passing then I guess that it is okay

langstream-core/src/main/java/ai/langstream/impl/common/BasicClusterRuntime.java
eolivelli: if all the tests are passing then I guess that it is okay

PR #622, opened 2023-10-19
CLI: cache github repository for faster deployments

nicoloboschi:
Currently every time the user specify a github repository for the application/secrets/instance we clone the entire git repository N times. If the user submit the deploy again, the repo is cloned from sctratch again. 99% of use cases are to download from this repo and it takes up to 10 seconds. 

Changes:
* Save git repository clone in ~/.langstream/ghrepos with the repo identifier. If the repo specified already exists, we just reset and fetch to the specified branch. **This is much faster than cloning the repo: ~3-4s vs 10~12s on my machine**
* In case the same command specify the same repository, the second one will be considered up to date and nothing will be done (neither clone or fetch)
* Added `--disable-local-repositories-cache` to fallback to the old behaviour
* In case of failures, the repo is wiped off and it restarts from scratch 

The downside of this solution is that the repository is stored "forever" in the user home but I don't think it's a real issue.
Another possible issue but very very unlikely is that two deployments happen in parallel and If at the same time the remote branch is updated. I don't think it's a real issue too

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/CLILogger.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/CLILogger.java
@@ -1,0 +1,36 @@
+package ai.langstream.cli;
+
+public interface CLILogger {
+    void log(Object message);
+
+    void error(Object message);
+
+    boolean isDebugEnabled();
+
+    void debug(Object message);
+
+
+    class SystemCliLogger implements CLILogger {
+        @Override
+        public void log(Object message) {
+            System.out.println(message);
+
+        }
+
+        @Override
+        public void error(Object message) {
+            System.err.println(message);
+        }
+
+        @Override
+        public boolean isDebugEnabled() {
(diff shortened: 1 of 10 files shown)


eolivelli:
Can we make this behavior configurable?

eolivelli (approved):
Lgtm

What happens in case of corrupted directory ?

langstream-cli/src/main/java/ai/langstream/cli/CLILogger.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        @Override
+        public void error(Object message) {
+            System.err.println(message);
+        }
+
+        @Override
+        public boolean isDebugEnabled() {
+            return true;
```
eolivelli, line 40: are we sure about this ?

eolivelli (approved):

PR #624, opened 2023-10-19
Add possibility to set retention policies in Pulsar

cbornet:


langstream-pulsar/src/main/java/ai/langstream/pulsar/PulsarClusterRuntimeConfiguration.java
```diff
@@ -15,20 +15,26 @@
...
 
-@Data
-@NoArgsConstructor
-@AllArgsConstructor
-public class PulsarClusterRuntimeConfiguration {
+public record PulsarClusterRuntimeConfiguration(
+        Map<String, Object> admin,
+        Map<String, Object> service,
+        Map<String, Object> authentication,
+        @JsonProperty("default-tenant") String defaultTenant,
```
nicoloboschi, line 25: this is clearly a breaking change, are we ok with that ? @eolivelli

cbornet: We talked about this with @eolivelli  .It's OK. Nobody could use Pulsar before.

nicoloboschi: I agree

PR #643, opened 2023-10-24
WebCrawler: use disk to store status if configured

nicoloboschi:
Changes:
* Added new parameter `state-storage`: `s3|disk` (default "s3") to the webcrawler.
* If disk is configured, a disk is automatically requested at runtime. the size is inferred from `resources.disk.size` if configured ,otherwise is 256M. The type is also inferred if configured
* Created implementation to store the status on local disk instead of s3

### DIFF
--- a/langstream-agents/langstream-agent-grpc/src/test/java/ai/langstream/agents/grpc/GrpcAgentSourceTest.java
+++ b/langstream-agents/langstream-agent-grpc/src/test/java/ai/langstream/agents/grpc/GrpcAgentSourceTest.java
@@ -252,11 +252,6 @@ public class GrpcAgentSourceTest {
         }
 
         @Override
-        public String getAgentId() {
-            return null;
-        }
-
-        @Override
         public TopicAdmin getTopicAdmin() {
             return null;
         }
--- a/langstream-agents/langstream-agent-grpc/src/test/java/ai/langstream/agents/grpc/GrpcAgentSourceTest.java
+++ b/langstream-agents/langstream-agent-grpc/src/test/java/ai/langstream/agents/grpc/GrpcAgentSourceTest.java
@@ -252,6 +252,11 @@ public class GrpcAgentSourceTest {
         }
 
         @Override
+        public String getAgentId() {
+            return null;
+        }
+
+        @Override
         public TopicAdmin getTopicAdmin() {
             return null;
         }
--- a/langstream-api/src/main/java/ai/langstream/api/runner/code/AgentContext.java
+++ b/langstream-api/src/main/java/ai/langstream/api/runner/code/AgentContext.java
@@ -36,6 +36,19 @@ public interface AgentContext {
 
     TopicProducer getTopicProducer();
 
+    /**
+     * Unique id of the agent relative to the application.
+     *
+     * @return
+     */
+    String getAgentId();
+
+    /**
+     * Unique id of the agent relative to the tenant. This is normally computed by
+     * {applicationId}-{agentId}
+     *
+     * @return
+     */
     String getGlobalAgentId();
 
     TopicAdmin getTopicAdmin();
--- a/langstream-api/src/main/java/ai/langstream/api/runner/code/AgentContext.java
+++ b/langstream-api/src/main/java/ai/langstream/api/runner/code/AgentContext.java
@@ -36,19 +36,6 @@ public interface AgentContext {
 
     TopicProducer getTopicProducer();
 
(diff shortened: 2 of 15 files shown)


eolivelli (changes requested):
we must provide an explicit configuration parameter to say that you want to use S3 or a disk attached to the agent.

`state-storage: "s3" | "agent-disk"`


in the case that the user configures "agent-disk" then the NodeProvider may automatically configure a disk, with the default storage type and something like 256MB of space

langstream-agents/langstream-agent-grpc/src/test/java/ai/langstream/agents/grpc/GrpcAgentSourceTest.java
```diff
@@ -251,6 +251,11 @@ public String getGlobalAgentId() {
             return null;
         }
 
+        @Override
+        public String getAgentId() {
```
eolivelli, line 255: how does this work for the "composite agent".
I think this is wrong.
The agent can use this.getAgentId()
we should not provide this method in the "AgentContext", that is shared among all the agents in a composite pipeline

langstream-api/src/main/java/ai/langstream/api/runner/code/AgentContext.java
```diff
@@ -36,6 +36,19 @@ public interface AgentContext {
 
     TopicProducer getTopicProducer();
 
+    /**
+     * Unique id of the agent relative to the application.
+     *
+     * @return
+     */
+    String getAgentId();
```
eolivelli, line 44: see my comment above, this method cannot be here

langstream-api/src/main/java/ai/langstream/api/runner/code/AgentContext.java
```diff
@@ -36,6 +36,19 @@ public interface AgentContext {
...
+
+    /**
+     * Unique id of the agent relative to the tenant. This is normally computed by
+     * {applicationId}-{agentId}
```
eolivelli, line 48: this is basically the id of the "executor", that is of the statefulset.
it is not about the "agent"

This comment is wrong

langstream-k8s-runtime/langstream-k8s-runtime-core/src/main/java/ai/langstream/runtime/impl/k8s/agents/WebCrawlerSourceAgentProvider.java
```diff
@@ -40,6 +46,42 @@ protected final ComponentType getComponentType(AgentConfiguration agentConfigura
...
+    @Override
+    protected DiskSpec computeDisk(
+            AgentConfiguration agentConfiguration,
+            Module module,
+            Pipeline pipeline,
+            ExecutionPlan physicalApplicationInstance,
+            ComputeClusterRuntime clusterRuntime,
+            StreamingClusterRuntime streamingClusterRuntime) {
+        final String stateStorage =
+                (String) agentConfiguration.getConfiguration().get("state-storage");
```
eolivelli, line 58: we should use ConfigurationUtils.getString here

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
```diff
@@ -347,6 +348,7 @@ private static void runJavaAgent(
                     AgentContext agentContext =
                             new SimpleAgentContext(
                                     agentId,
+                                    configuration.agent(),
```
eolivelli, line 351: this is wrong, in case of the "composite agent" this doesn't make sense, or maybe it is the id of the first agent in the group

langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
```diff
@@ -1000,7 +1004,8 @@ public SimpleAgentContext(
             this.consumer = consumer;
             this.producer = producer;
             this.topicAdmin = topicAdmin;
-            this.agentId = agentId;
+            this.globalAgentId = globalAgentId;
+            this.agentId = agentSpec.agentId();
```
eolivelli, line 1008: see comments above

eolivelli (changes requested):

eolivelli (approved):
LGTM

PR #644, opened 2023-10-24
[agents] Add support for using Camel components as sources

eolivelli:
This is still a POC.

Summary:
- add a camel-source agent that allows you to use any Camel component as a Source for LangStream
- add a sample application that uses The GitHub camel connector
- it is possible to choose an header coming from the Camel message and set is as key of the output record
- the camel-source implements the Camel exchange protocol and it acks the messages completely processed by the pipeline, and it is also notifying failures

With Camel components LangStream fully support at-least once semantics 

Notes:
we are not bundling any Camel connectors, you have to add the connector as a dependency, as it is shown in the sample application. The "dependency" mechanism that you use for JDBC and for Kafka Connect works also for Camel


eolivelli:
I have updated the sample application

eolivelli:
I realised that I have to also encode the headers

eolivelli:
test

eolivelli:
test3

eolivelli:
test 5

langstream-agents/langstream-agent-camel/src/main/java/ai/langstream/agents/camel/CamelSource.java
```diff
@@ -0,0 +1,249 @@
...
+        int maxRecords = ConfigurationUtils.getInt("max-buffered-records", 100, configuration);
+        this.keyHeader = ConfigurationUtils.getString("key-header", "", configuration);
+        Map<String, Object> componentOptions =
+                ConfigurationUtils.getMap("component-options", Map.of(), configuration);
+        if (componentOptions != null) {
+            for (Map.Entry<String, Object> entry : componentOptions.entrySet()) {
+                String paramName = entry.getKey();
+                Object value = entry.getValue();
+                if (value != null) {
+                    if (componentUri.contains("?")) {
```
cbornet, line 182: Could this be done with `java.net.URI` ?

eolivelli: I don't know if Camel components URI are real URIs, IIRC there are some tricks, like RAW(xxx)

PR #646, opened 2023-10-25
[examples] Add a chatbot built with LangChain and Astra DB

eolivelli:
This is the first iteration of the sample application about how to build a chat bot using LangChain and the RAG pattern.

Summary:
- the code is inspired by https://github.com/langchain-ai/chat-langchain
- the application uses DataStax Astra DB as vector database
- OpenAI is used as LLM for completions and for computing the text embeddings
- as in the original chat application, the chat history is to be handled on the client side

### DIFF
--- a/examples/applications/langchain-chat/python/langchain_chat.py
+++ b/examples/applications/langchain-chat/python/langchain_chat.py
@@ -42,7 +42,6 @@ from langchain.schema.runnable import (
 from langchain.vectorstores import Weaviate
 from pydantic import BaseModel
 
-
 RESPONSE_TEMPLATE = """\
 You are an expert programmer and problem-solver, tasked with answering any question \
 about Langchain.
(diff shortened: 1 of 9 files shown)


examples/applications/langchain-chat/python/langchain_chat.py
```diff
@@ -0,0 +1,242 @@
...
+            *,
+            callbacks: Callbacks = None,
+            tags: Optional[List[str]] = None,
+            metadata: Optional[Dict[str, Any]] = None,
+            **kwargs: Any,
+    ) -> List[Document]:
+        return [Document(page_content="foo"), Document(page_content="bar")]
+
+
+def get_retriever(embeddingsModel) -> BaseRetriever:
```
cbornet, line 113: This code doesn't seem to be used

PR #647, opened 2023-10-25
CLI UI: start on port 8092 by default

### LINKED ISSUE #641: CLI UI: option for using a well known port
Currently the UI starts up at some random port which is hard to track if you're in a WSL environment. 
We should add a parameter for forcing the port. Both for apps ui and docker run commands

Also in docker run the port is not even logged

nicoloboschi:
Fixes #641 

Changes:
* By default apps ui and docker run start the UI on port 8092
* There's a option to change the port. 0 means ephimeral (like before). `-p` for apps ui and `-up` for docker run

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
@@ -15,6 +15,7 @@
  */
 package ai.langstream.cli.commands.applications;
 
+import ai.langstream.cli.CLILogger;
 import ai.langstream.cli.NamedProfile;
 import ai.langstream.cli.api.model.Gateways;
 import io.undertow.Handlers;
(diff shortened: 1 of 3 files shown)


langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
```diff
@@ -56,6 +57,12 @@ public class UIAppCmd extends BaseApplicationCmd {
     @CommandLine.Parameters(description = "Application ID")
     private String applicationId;
 
+    @CommandLine.Option(
+            names = {"-p", "--port"},
```
eolivelli, line 61: what about "--ui-port" ? "--port" is too generic

eolivelli (changes requested):

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
nicoloboschi: this command is litereally 
```
apps iu <appid>
```
I don't think it's confusing

In fact, in the docker run one, I used `--ui-port`

eolivelli (approved):
LGTM

PR #652, opened 2023-10-26
[python] Add a CLI command to import pip requirements.txt

eolivelli:
Summary:

This path adds a new CLI command to process the requrements.txt file using the same version of the OS and of Python that run in production, on Linux/Intel.

This is very useful in order to ensure that the dependencies that you import match correctly the production enviroment

`langstream python load-pip-requirements -app examples/applications/python-processor-exclamation`

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/python/LoadPythonDependenciesCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/python/LoadPythonDependenciesCmd.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


langstream-cli/src/main/java/ai/langstream/cli/commands/python/LoadPythonDependenciesCmd.java
```diff
@@ -0,0 +1,213 @@
...
+    @CommandLine.Option(
+            names = {"--langstream-runtime-docker-image"},
+            description = "Docker image of the LangStream runtime to use")
+    private String dockerImageName;
+
+    @Override
+    @SneakyThrows
+    public void run() {
+
+        if (dockerImageVersion != null && dockerImageVersion.endsWith("-SNAPSHOT")) {
```
nicoloboschi, line 70: can we move it to a common function? is the same logic of docker run, right ?

eolivelli: good point, done

PR #654, opened 2023-10-26
[python] Add support for hot reload of python code in docker mode

eolivelli:
Summary:

When you are working locally, add the ability to automatically reload python code without restarting the container

In "docker run" mode the cli start watching on the python files.
In case of any python file that has changed the python agent is restarted, picking up the new version of the code.

The is a new flag "--watch-files=false" to disable the feature.

The watcher starts only if there is a "python" directory in the application

Follow up:
- if there is a syntax error the Python process crashes, so the docker container exists
- add test cases

### DIFF
--- a/langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
+++ b/langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
@@ -102,7 +102,7 @@ public class LocalRunApplicationCmd extends BaseDockerCmd {
 
     @CommandLine.Option(
             names = {"--watch-files"},
-            description = "Start the UI")
+            description = "Watch files and apply the changes automatically (only Python files)")
     private boolean watchFiles = true;
 
     @CommandLine.Option(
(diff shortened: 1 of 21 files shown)


langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/api/AgentInfoServlet.java
```diff
@@ -20,19 +20,38 @@
...
     @Override
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
         resp.setContentType("application/json");
-        MAPPER.writeValue(resp.getOutputStream(), agentInfo.serveWorkerStatus());
+        MAPPER.writeValue(resp.getOutputStream(), agentAPIController.serveWorkerStatus());
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+        resp.setContentType("application/json");
```
eolivelli, line 43: This is currently not used

langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
```diff
@@ -96,6 +100,11 @@ public class LocalRunApplicationCmd extends BaseDockerCmd {
             description = "Start the UI")
     private boolean startUI = true;
 
+    @CommandLine.Option(
+            names = {"--watch-files"},
+            description = "Start the UI")
```
nicoloboschi, line 105: to change

langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
```diff
@@ -380,26 +393,75 @@ private void executeOnDocker(
...
+                    }
+                },
+                watcher);
+    }
+
+    private static void restartAgents() {
+        HttpURLConnection urlConnection = null;
+        try {
+            URL url = new URL("http://localhost:8790/commands/restart");
+            System.out.println("Calling " + url + " to restart the agents");
```
nicoloboschi, line 454: can you use the cli logger?

eolivelli: sure

PR #656, opened 2023-10-27
[docker] Do not close classloaders, to prevent spammy ClassNotFoundExceptions

eolivelli:
When the docker container is existing, usually due to an error in the configuration, we close the classloaders and this triggers a lot of spammy ClassNotFound errors that are misleading, because users think that there is a bug somewhere in LangStream.

The trick here is not not close them: the JMV is existing, there is no risk of leaking resources.
Also the tmp directory of the container is ephemeral: we are not leaving dangling temporary files

### DIFF
--- a/langstream-core/src/main/java/ai/langstream/impl/nar/NarFileHandler.java
+++ b/langstream-core/src/main/java/ai/langstream/impl/nar/NarFileHandler.java
@@ -47,6 +47,13 @@ public class NarFileHandler
                 AssetManagerRegistry.AssetManagerPackageLoader,
                 TopicConnectionsRuntimeRegistry.TopicConnectionsPackageLoader {
 
+    private static final boolean CLOSE_CLASSLOADERS =
+            Boolean.parseBoolean(System.getProperty("langstream.nar.closeClassloaders", "true"));
+
+    static {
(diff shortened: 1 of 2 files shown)


langstream-core/src/main/java/ai/langstream/impl/nar/NarFileHandler.java
```diff
@@ -47,6 +47,13 @@ public class NarFileHandler
                 AssetManagerRegistry.AssetManagerPackageLoader,
                 TopicConnectionsRuntimeRegistry.TopicConnectionsPackageLoader {
 
+    private static final boolean CLOSE_CLASSLOADERS =
+            Boolean.parseBoolean(System.getProperty("langstream.nar.closeClassloaders", "true"));
+
+    static {
+        log.info("langstream.nar.close-classloaders = {}", CLOSE_CLASSLOADERS);
```
nicoloboschi, line 54: please align to the system property name

nicoloboschi (changes requested):

langstream-core/src/main/java/ai/langstream/impl/nar/NarFileHandler.java
eolivelli: good catch

nicoloboschi (approved):

PR #660, opened 2023-10-27
API gateway: cache topic producer for the same gateway

nicoloboschi:
Changes:
* We use Guava cache to keep producer by gateway. The cache is LRU and every access refresh the ttl. 
* The max size is configurable (max 100 across all tenants). The cache can be disabled

### DIFF
--- a/langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/LRUTopicProducerCache.java
+++ b/langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/LRUTopicProducerCache.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.langstream.apigateway.gateways;
+
+import ai.langstream.api.runner.code.Record;
+import ai.langstream.api.runner.topics.TopicProducer;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalNotification;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class LRUTopicProducerCache implements TopicProducerCache {
+
+    private static class SharedTopicProducer implements TopicProducer {
+        private TopicProducer producer;
+        private volatile int referenceCount;
+        private volatile boolean cached = true;
+
+        public SharedTopicProducer(TopicProducer producer) {
+            this.producer = producer;
+        }
+
+        public boolean acquire() {
+            synchronized (this) {
+                if (producer == null) {
+                    return false;
+                }
+                referenceCount++;
+                return true;
+            }
+        }
+
+        public void removedFromCache() {
+            synchronized (this) {
+                cached = false;
+                if (referenceCount == 0) {
+                    producer.close();
+                    producer = null;
+                }
+            }
+        }
+
+        @Override
+        public void start() {
+            producer.start();
+        }
+
+        @Override
+        public void close() {
+            synchronized (this) {
+                referenceCount--;
+                if (referenceCount == 0 && !cached) {
+                    producer.close();
+                    producer = null;
+                }
+            }
+        }
+
+        @Override
+        public CompletableFuture<?> write(Record record) {
+            return producer.write(record);
+        }
+
+        @Override
+        public Object getNativeProducer() {
+            return producer.getNativeProducer();
+        }
+
+        @Override
+        public Object getInfo() {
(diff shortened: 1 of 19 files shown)


langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/LRUTopicProducerCache.java
```diff
@@ -0,0 +1,118 @@
...
+                        .removalListener(
+                                (RemovalNotification<Key, SharedTopicProducer> notification) -> {
+                                    SharedTopicProducer resource = notification.getValue();
+                                    resource.removedFromCache();
+                                })
+                        .build();
+    }
+
+    @Override
+    public synchronized TopicProducer getOrCreate(
```
eolivelli, line 107: synchronized is not needed here, the cache is already threadsafe, also this would be a point in which many thread will be blocked while on producer is being created

langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/LRUTopicProducerCache.java
```diff
@@ -0,0 +1,118 @@
...
+            TopicProducerCache.Key key, Supplier<TopicProducer> topicProducerSupplier) {
+        try {
+            final SharedTopicProducer sharedTopicProducer =
+                    cache.get(key, () -> new SharedTopicProducer(topicProducerSupplier.get()));
+            sharedTopicProducer.acquire();
```
eolivelli, line 112: this is a common pitfall

the "acquire" should be performed in the constructor, because between this line and the previous line it is possible that the producer has been released if the count is zero (t is an edge case, but it usually happens).

The tricky thing is that you have to call acquire if there is already an object.

eolivelli: Also the producer must be started only once, so you have to start it when you have created it
and the cache should not return the producer until it is fully initialised (started) with success
and if the start operation fails it must not be added to the cache

langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/ProduceGateway.java
```diff
@@ -96,33 +100,38 @@ public void start(
...
                         .asTopicConnectionsRuntime();
 
         topicConnectionsRuntime.init(streamingCluster);
 
-        producer =
+        final TopicProducer topicProducer =
                 topicConnectionsRuntime.createProducer(
                         null, streamingCluster, Map.of("topic", topic));
-        producer.start();
+        topicProducer.start();
```
eolivelli, line 132: this is not to be executed here, see the comments above

langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/TopicProducerCache.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package ai.langstream.apigateway.gateways;
+
+import ai.langstream.api.runner.topics.TopicProducer;
+import java.util.function.Supplier;
+
+public interface TopicProducerCache {
+    record Key(String tenant, String application, String gatewayId) {}
+
+    TopicProducer getOrCreate(Key key, Supplier<TopicProducer> topicProducerSupplier);
```
eolivelli, line 24: this operation may fail, it is better to add a "throws" clause. The risk is to forget possible failures (that will happen for instance if the broker is down)

langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/TopicProducerCacheFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+@Configuration
+public class TopicProducerCacheFactory {
+
+    @Bean
+    public TopicProducerCache topicProducerCache(TopicProperties topicProperties) {
+        if (topicProperties.isProducersCacheEnabled()) {
+            return new LRUTopicProducerCache(topicProperties.getProducersCacheSize());
+        } else {
+            return (key, topicProducerSupplier) -> topicProducerSupplier.get();
```
eolivelli, line 30: This is tricker then expected, see the comments above
the producer that is returned should be already ready to accept writes (started)

langstream-api-gateway/src/main/resources/application.properties
```diff
@@ -29,4 +29,7 @@ spring.jackson.serialization.order-map-entries-by-keys=true
 application.storage.apps.type=kubernetes
 application.storage.apps.configuration.namespaceprefix=langstream-
 
-application.gateways.code.path=/app/agents
\ No newline at end of file
+application.gateways.code.path=/app/agents
+
+application.topics.producers-cache-enabled=true
+application.topics.producers-cache-size=100
```
eolivelli, line 35: we should have metrics for this, otherwise it is hard to understand if the system is close to the threshold.
when you hit the threshold we stop caching the system will have very bad performances (latency)

eolivelli (changes requested):

langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/LRUTopicProducerCache.java
nicoloboschi: the producer is started in during the load in the cache and every producer in the cache is already started.

nicoloboschi: the acquire must take into condiseration the edge case you mentioned, right. I did it in another manner. When you call acquire, if the producer has been actually closed (removed from cache for example), we retry and that will trigger the constructor to be called

langstream-api-gateway/src/main/java/ai/langstream/apigateway/gateways/TopicProducerCacheFactory.java
nicoloboschi: yes it is

eolivelli (approved):

PR #663, opened 2023-10-27
New gateway 'service' - part 1

nicoloboschi:
Changes:
* added new gateway type `service`, very similar to the chat one but only for http. The user must send a input and the gateways wait for a message from the output topic

```
- id: g4
  type: service
  service-options:
    input-topic: input
    output-topic: output
    headers:
    - value-from-parameters: v1
```
* The http endpoint is 
 http://xxx/api/gateways/service/tenant1/application1/g4
* The body format is the same as the produce one 
```
{
 "key": "k1", 
  "value": "translate .."
  headers: {..}
}
```


langstream-api-gateway/src/main/java/ai/langstream/apigateway/http/GatewayResource.java
```diff
@@ -83,19 +88,102 @@ ProduceResponse produce(
...
+    @PostMapping(
+            value = "/service/{tenant}/{application}/{gateway}",
+            consumes = MediaType.APPLICATION_JSON_VALUE)
+    void service(
+            WebRequest request,
+            HttpServletResponse response,
+            @NotBlank @PathVariable("tenant") String tenant,
+            @NotBlank @PathVariable("application") String application,
+            @NotBlank @PathVariable("gateway") String gateway,
+            @RequestBody ProduceRequest produceRequest)
```
eolivelli, line 120: where do you set the timeout ?
I guess that we should give the client the ability to set the timeout for waiting for the message to arrive

langstream-api-gateway/src/main/java/ai/langstream/apigateway/http/GatewayResource.java
```diff
@@ -83,19 +88,102 @@ ProduceResponse produce(
...
+                        headers,
+                        new ProduceGateway.ProduceGatewayRequestValidator());
+        final AuthenticatedGatewayRequestContext authContext;
+        try {
+            authContext = gatewayRequestHandler.authenticate(context);
+        } catch (GatewayRequestHandler.AuthFailedException e) {
+            throw new ResponseStatusException(HttpStatus.UNAUTHORIZED, e.getMessage());
+        }
+
+        try (final ConsumeGateway consumeGateway =
```
eolivelli, line 141: we have to follow up with a cache for this

eolivelli (approved):
LGTM

PR #665, opened 2023-10-29
[python] Add new agent type python-service and add support for running LangServe applications

eolivelli:
Summary:
- add a new component type "SERVICE"
- a Service doesn't use streaming to interact with external services
- the service is supposed to open some port on the pod and serve an HTTP service (no need to support TLS at the moment)
- initially this port is fixed and it must be 8000, in the future we can make it configurable
- add a new agent "python-service" that allows you to deploy a python application without consuming from a topic or producing messages
- add LangServe, FastAPI and Uvicorn to the bundled Python libraries

when you are running in docker the auto-reload feature works out of the box, as soon as you change your Python files the Python daemon is restarted

It is expected that the langstream gateway will act as a proxy to the service, adding:
- TLS support for public networks
- Authentication

Running Python application on Kubernetes

LangStream agents are already deployed as a StatefulSet (with persistent disks) and an headless Service.
Deploying an application here allows you to easily leverage all the Kubernetes features, for instance:
- resource management (you can set set size of the machines, CPU, memory...)
- scalability (you can set the number of replicats)
- region awareness
- ... 

Bonus feature added by this PR:

When you declare a Python agent you can easily pass environment variables from secrets:

```
module: default
name: "Basic LangServe application"
pipeline:
  - name: "Start LangServe"
    id: agent1
    type: "python-service"
    configuration:
      className: example.ChatBotService
      environment:
        - key: "OPENAI_API_KEY"
          value: "${secrets.open-ai.access-key}"
        - key: "LANGCHAIN_TRACING_V2"
          value: "true"
        - key: "LANGCHAIN_ENDPOINT"
          value: "${ secrets.lang-smith.api-url }"
        - key: "LANGCHAIN_API_KEY"
          value: "${ secrets.lang-smith.api-key }"          
```

This PR adds a sample application about how to integrate also with LangSmith
image


docker/build.sh
```diff
@@ -32,9 +33,8 @@ common_flags="-DskipTests -PskipPython -Dlicense.skip -Dspotless.skip -ntp"
 
 build_docker_image() {
   module=$1
-  clean="${2:-true}"
   clean_cmd=""
-  if [ "$clean" == "true" ]; then
```
nicoloboschi, line 37: please revert, this was done on purpose, on some modules is better to clean before building

langstream-k8s-deployer/langstream-k8s-deployer-core/src/main/java/ai/langstream/deployer/k8s/agents/AgentResourcesFactory.java
```diff
@@ -223,9 +228,15 @@ public static StatefulSet generateStatefulSet(GenerateStatefulsetParams params)
                         .withTerminationMessagePolicy("FallbackToLogsOnError")
                         .build();
 
-        final ContainerPort port =
+        final ContainerPort portHttp =
                 new ContainerPortBuilder().withName("http").withContainerPort(8080).build();
 
+        final ContainerPort portCustomServiceHttp =
```
nicoloboschi, line 234: always ? shouldn't we do it only if the application has a python-service ?

nicoloboschi (changes requested):

docker/build.sh
eolivelli: reverted

langstream-k8s-deployer/langstream-k8s-deployer-core/src/main/java/ai/langstream/deployer/k8s/agents/AgentResourcesFactory.java
eolivelli: we have to tweak the CRD (again) in order to do it.
I am not sure it is worth.

a note:
it is not only for "python-service" but for any agent with componentType=SERVICE.

PR #670, opened 2023-10-30
CLI: add .env configuration file

nicoloboschi:
Changes:
* Added possibility add an ENV file that is loaded before the actual java command. the file must be in the bin directory and must be named `langstream.env`
* Docker image by default is retrieved from LANGSTREAM_RUNTIME_DOCKER_IMAGE env


langstream-cli/src/main/java/ai/langstream/cli/commands/docker/LocalRunApplicationCmd.java
```diff
@@ -148,7 +148,8 @@ public class LocalRunApplicationCmd extends BaseDockerCmd {
 
     @CommandLine.Option(
             names = {"--langstream-runtime-docker-image"},
-            description = "Docker image of the LangStream runtime to use")
+            description = "Docker image of the LangStream runtime to use",
```
eolivelli, line 151: There is also the python command that processes requirements.txt that uses a docker image

eolivelli (changes requested):

PR #673, opened 2023-10-31
[agents] Add Agent to call LangServe (langserve-invoke)

eolivelli:
Summary:
- add a new agent "langserve-invoke" that is able to make a call to LangServe
- support both "invoke" and "stream" endpoints
- in case of "stream" endpoints you can configure the "stream-to-topic" property, like in the ai-chat-completion agent

When doing streaming answers the chunking is performed the same way as the ai-chat-completion agent, and we are adding the same properties to the messages. This way the sample JS applications still work out of the box.

There is a sample application included with this PR.

Sample pipeline:

```
pipeline:
  - type: "langserve-invoke"
    input: input-topic
    output: output-topic
    id: step1
    configuration:
      output-field: value.answer
      stream-to-topic: streaming-answers-topic
      stream-response-field: value
      min-chunks-per-message: 10
      debug: false
      method: POST
      allow-redirects: true
      handle-cookies: false
      url: "http://host.docker.internal:8000/chain/stream"
      fields:
        - name: topic
          expression: "value"
```


nicoloboschi (approved):
LGTM

PR #675, opened 2023-10-31
Fix WebCrawler state-storage default value

nicoloboschi:
Error (only on main branch)

```
Cannot invoke \"String.equals(Object)\" because \"stateStorage\" is null
```


eolivelli (approved):
LGTM

PR #676, opened 2023-11-01
[langserve-invoke] Extract the LangServe Java Client code and Allow to parse simple strings and not only JSON

eolivelli:
Summary:
- handle String content
- extract the code of the Java LangServeClient
- add tests


nicoloboschi (approved):
LGTM
adding a unit test would be better

PR #682, opened 2023-11-02
[cli]] updates for multi-os browser for langstream UI

mark878:
main issue was to figure out that the thread needs to wait a bit after processBuilder is called to start the browser and waitfor() can't be used because it doesn't return until the started process completes.

I only tested this out is WSL.  Needs to be tested on standard linux, macos and windows.

Added logic to determine OS and spawn a browser process based on the OS


mark878:
updated the test(s)

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
```diff
@@ -58,17 +61,20 @@ public class UIAppCmd extends BaseApplicationCmd {
     private String applicationId;
 
     @CommandLine.Option(
-            names = {"-p", "--port"},
-            description = "Port for the local webserver and UI. If 0, a random port will be used. ",
-            defaultValue = "8092")
+        names = {
```
eolivelli, line 64: this file has been reformatted

please run this command to make it complaint with the code style

mvn generate-sources

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
```diff
@@ -30,7 +30,10 @@
 import io.undertow.server.handlers.resource.ResourceHandler;
 import io.undertow.util.Headers;
 import io.undertow.util.HttpString;
+import java.awt.Desktop;
```
eolivelli, line 33: this is a great idea !

I think that Desktop.browse is meant to be compatible with all the OSs
https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/awt/Desktop.html#browse(java.net.URI)

I have tested it on Mac and it works like a charm

Can you please try to use it also on your OS ?

eolivelli (changes requested):

langstream-cli/src/test/java/ai/langstream/cli/commands/applications/UIAppCmdTest.java
```diff
@@ -24,8 +25,12 @@ class UIAppCmdTest {
     @Test
     void openBrowser() {
         // ok
-        UIAppCmd.openBrowserAtPort("echo", 9029);
+	String command = "/usr/bin/echo";
+	boolean Result = UIAppCmd.checkAndLaunch( command, 80);
```
eolivelli, line 29: I think that we can remove checkAndLaunch given your findings on Desktop.browse()
at that point we can drop this test.

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
```diff
@@ -119,91 +125,128 @@ public static Undertow startServer(
...
+            } catch (IOException ioException) {
+                return false;
+            }
+        } else {
+            return false;
         }
     }
 
+    static boolean openBrowserAtPort(String URL, int port) {
+        String os = System.getProperty("os.name").toLowerCase();
```
nicoloboschi, line 217: It's better to use commons-lang3 
> SystemUtils.IS_OS_MAC

nicoloboschi (changes requested):
Good work !

langstream-cli/src/test/java/ai/langstream/cli/commands/applications/UIAppCmdTest.java
mark878: Desktop.browse() didn't work in WSL linux.  Or rather it didn't when I tried it before.  That's why I went to xdg-open. 
 I can use Desktop.browse() to open the browser on a mac.  Will rewrite for mac and also try it again on WSL

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
mark878: I spent some time trying to figure what style to use.  Got educated on java style. thx will do.  BTW, I do everything at the command-line.  I use to use jetbrains -probably setup it up at some point.

mark878: done

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
mark878: It doesn't work in WSL linux.  I'll modify the code to launch if in a mac. For WSL linux, I tried it earlier. The WSL error was something like "can't launch headless".  Easy to try though.  Will try again.

mark878: Tried this.  At least I got a status message in the logs:
```
received response: (GET http://localhost:8090/api/applications/default/test?stats=false) 200
Starting UI at http://localhost:8092
Operating system identified as: linux
Could not Start browser.  Either add the proper command to your OS (open for mac or xdg-open for linux) or start a browser manually at http://localhost:8092
```

code change:
```

    static boolean openBrowserAtPort(String URL, int port) {
        String os = System.getProperty("os.name").toLowerCase();
        String openCommand = "";
        if (os.indexOf("win") >= 0
                || os.indexOf("nux") >= 0
                || os.indexOf("nix") >= 0
                || os.indexOf("nux") >= 0) {
            Desktop desktop = Desktop.getDesktop();
            try {
                desktop.browse(new URI(URL + port));
            } catch (Exception e) {
                return false;
            }
            return true;
        } else if (os.indexOf("nix") >= 0 || os.indexOf("nux") >= 0) {
            openCommand = "/usr/bin/xdg-open";
            return checkAndLaunch(openCommand, port);
        }
        return false;
```
    }

mark878: I will, however, have desktop browser to open the browser on a mac

langstream-cli/src/test/java/ai/langstream/cli/commands/applications/UIAppCmdTest.java
mark878: left checkAndLaunch in as WSL as Desktop.browse() fails on WSL linux (ubuntu 22.04).  Maybe it'll get fixed eventually.

eolivelli (changes requested):

langstream-cli/src/test/java/ai/langstream/cli/commands/applications/UIAppCmdTest.java
```diff
@@ -15,17 +15,41 @@
...
+
+        Files.setPosixFilePermissions(
+                path, Set.of(PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_EXECUTE));
+    }
+
     @Test
-    void openBrowser() {
+    void openBrowser() throws IOException {
+        // write test file
+        String fname = "/tmp/testfile";
```
eolivelli, line 42: This test cannot pass on a Window machine.

I think that we can remove it.
Also, it is creating a fixed path, so you cannot run the same test in parallel (not a big deal, but it worth mentioning it)

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
```diff
@@ -176,19 +179,55 @@ public static Undertow startServer(
...
+        String openCommand = "";
+        if (SystemUtils.IS_OS_MAC || SystemUtils.IS_OS_WINDOWS) {
+            Desktop desktop = Desktop.getDesktop();
+            try {
+                desktop.browse(new URI(URL + port));
+            } catch (Exception e) {
+                return false;
+            }
+            return true;
+        } else if (SystemUtils.IS_OS_LINUX) {
```
eolivelli, line 226: doesn't Desktop.browse() work on Linux ?

eolivelli (changes requested):

langstream-cli/src/main/java/ai/langstream/cli/commands/applications/UIAppCmd.java
mark878: not in WSL.  Browser window doesn't open becasuse desktop.browse throws an execption.  

However, I can manually open a browser to the UI URL

eolivelli (approved):
LGTM

PR #683, opened 2023-11-03
Fix service gateway with topics

nicoloboschi:
Changes:
* Invert input and output topic as expected
* add fixed header `langstream-service-request-id` to ensure getting the response to the previously sent message
* Http gateways now by default expects a plain/text that will be part the message value. By passing Content-Type application/json you can still pass the entire body 
* improved tests


eolivelli (approved):
LGTM

PR #690, opened 2023-11-06
Complete python gRPC requests when closing

cbornet:


eolivelli (approved):
+1

did you try the reload feature ?
I hope that this change doesn't break it.
Unfortunately it is to be tested manually, in order to see how it behaves for the user.

if we block waiting for the current response to complete and the application is broken then we don't want it to hang

eolivelli:
Also it seems that you have to rebase on top of main branch

cbornet:
I rebased and changed a bit the code.
The PR doesn't touch the `restart` procedure. It only affects the `close` and the tests.

cbornet:
I tested the restart manually. Works fine.

PR #692, opened 2023-11-06
[python agents] Allow to access resources configured in configuration.yaml

eolivelli:
Summary:
- in python agents LangStream is now passing all the "Resources" to the configuration of the agent
- add support for using database-id in AstraDB datasources

With this change you can declare the access to DataSources/Vector Database and to AI service providers only once in your application, then in Python you can use the values.
This way there is no need to add boilerplate to map the access keys in the pipeline files.

In python now you can access resources and globals:

- config["resources"]
- config["globals"]

### DIFF
--- a/examples/applications/langchain-chat/python/langchain_chat.py
+++ b/examples/applications/langchain-chat/python/langchain_chat.py
@@ -201,12 +201,13 @@ def create_chain(
 class LangChainChat(Processor):
 
     def init(self, config):
-        self.openai_key = config.get("openai-key", "")
-        self.astra_db_token = config.get("astra-db-token", "")
-        self.astra_db_id = config.get("astra-db-id", "")
-        self.astra_keyspace = config.get("astra-db-keyspace", "")
-        self.astra_table_name = config.get("astra-db-table", "")
 
+        # the values are configured in the resources section in configuration.yaml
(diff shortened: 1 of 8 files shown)


examples/applications/langchain-chat/python/langchain_chat.py
```diff
@@ -201,12 +201,13 @@ def create_chain(
...
 
     def init(self, config):
-        self.openai_key = config.get("openai-key", "")
-        self.astra_db_token = config.get("astra-db-token", "")
-        self.astra_db_id = config.get("astra-db-id", "")
-        self.astra_keyspace = config.get("astra-db-keyspace", "")
-        self.astra_table_name = config.get("astra-db-table", "")
 
+        # the values are configured in the resources section in configuration.yaml
+        self.openai_key = config.resources.openai.get("access-key", "");
```
cbornet, line 206: I'm surprised you can reference `config.resources`. Shouldn't it be `config["resources"]` ?

cbornet (changes requested):
Just a comment on the langchain example. Otherwise LGTM.

examples/applications/langchain-chat/python/langchain_chat.py
eolivelli: you are right

cbornet (approved):

Issue #698, opened 2023-11-07
mini-langstream start hangs after HerdDB - service langstream/langstream-control-plane has no node port


Gagravarr:
I'm trying to follow https://docs.langstream.ai/installation/get-started-minikube to get langstream running on minikube. I installed the dependencies, ran the get shell script, then tried to start. However, `mini-langstream start` hangs after `HerdDB: ✅`

It seems to be stuck in the `start_port_forward` step - https://github.com/LangStream/langstream/blob/main/mini-langstream/mini-langstream#L227

In `/tmp/` I can see the file it is stuck grep'ing, which shows

```
* service langstream/langstream-control-plane has no node port
```

There's no error handling in `start_port_forward` so the script hangs forever if there's no URL returned...

It looks like the services have been defined without external ports

```
mini-langstream kubectl get svc
NAME                       TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
langstream-api-gateway     ClusterIP   10.106.115.58    <none>        8091/TCP   48m
langstream-control-plane   ClusterIP   10.111.103.230   <none>        8090/TCP   48m
```

Not sure if there's a missing setup step in the docs, a missing config step in the `mini-langstream` script, or a bug in the helm script?

eolivelli:
Are you on Linux?

@nicoloboschi can you please take a look?

Gagravarr:
Yup, linux, running `Ubuntu 22.04.3 LTS` (newly setup Azure VM today)

Gagravarr:
In case it helps, the versions of everything I'm running are:
```
*Docker:* Docker version 24.0.5, build 24.0.5-0ubuntu1~22.04.1
*Minikube:* minikube version: v1.31.2
*Helm:* version.BuildInfo{Version:"v3.13.1", GitCommit:"3547a4b5bf5edb5478ce352e18858d8a552a4110", GitTreeState:"clean", GoVersion:"go1.20.8"}
*KubeCtl:* Client Version: v1.28.3, Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
*Ubuntu:* Ubuntu 22.04.3 LTS
*LangStream:* LangStream CLI 0.4.2
*Mini LangStream:* 0.4.2
```

Running the wrapped minikube we see no node ports setup for the langstream services, but ingress is enabled in general
```
KUBECONFIG=.mini-langstream/data/kube-config minikube --profile mini-langstream service list
[...]

PR #700, opened 2023-11-07
MiniLangstream should error out if no NodePort exists for a service (#698)

Gagravarr:
Add error handling to the shell script function `start_port_forward` to avoid the current hanging if no *node port* is defined for the service.

Doesn't attempt any alternate forwarding (eg kubectl), just gives a moderately helpful message and errors out.

### DIFF
--- a/mini-langstream/mini-langstream
+++ b/mini-langstream/mini-langstream
@@ -223,10 +223,15 @@ start_port_forward() {
   local service=$1
   local file=$(mktemp)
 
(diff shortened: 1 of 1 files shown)


eolivelli (approved):
Lgtm

@nicoloboschi can you please review here?

eolivelli:
Please fill in https://cla.datastax.com then we can merge your patch.
Thanks

nicoloboschi (approved):

PR #714, opened 2023-11-10
Apps: implement force deletion and improve status

nicoloboschi:
Changes:
* If the deployer/setup job during deploy or deletion fails, we now report the error in the application status
* There's a new option in the deletion `-f`, to skip cleanup errors and go ahead with the application deletion

Implementation:
* There's a new field in the application crd that handles wether the application can skip the cleanup or not in case of failures (only 1 attempt).
* The deletion process totally changes:
  * Before: the control plane deletes the app CRD and the finalizer is triggered in the operator. This triggers the cleanup and in case of failures, nothing can be done. 
  * Now: the control plane puts a flag "marked for deletion" in the crd specs. In this way it's then possible to force the deletion of the app after the app failed to be deleted. This was not possible before because once you delete a custom resource, you cannot change his config. So now the operator reacts to this new flag by starting the deletion process, handling errors reading the new skipErrors flag. Once done, it finally deletes the custom resource. The finalizer is still there to verify the cleanup ran and to update the tenant resource usage, but now it's much more lightweight.

### DIFF
--- a/helm/crds/applications.langstream.ai-v1.yml
+++ b/helm/crds/applications.langstream.ai-v1.yml
@@ -27,6 +27,8 @@ spec:
                 type: string
               imagePullPolicy:
                 type: string
+              options:
+                type: string
               tenant:
                 type: string
             required:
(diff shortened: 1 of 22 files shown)


eolivelli (approved):
LGTM

very nice !

PR #717, opened 2023-11-13
Added build instructions to readme

devinbost:
Added additional requirements for building LangStream and added some simple instructions for fixing CI style issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,19 @@ Requirements for building the project:
 * Java 17
 * Git
 * Python 3.11+ and PIP
+* Maven 3.8+
+
+Additionally, you will need to install into your Python env:
+* Black
+* Ruff
+* Build
+
+You can verify your requirements and run your first build like this:
+
+```bash
+mvn clean install -DskipTests
+mvn generate-sources
+```
 
 If you want to test local code changes, you can use `mini-langstream`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,6 +180,19 @@ Requirements for building the project:
...
+Additionally, you will need to install into your Python env:
+* Black
+* Ruff
+* Build
+
+You can verify your requirements and run your first build like this:
+
+```bash
+mvn clean install -DskipTests
+mvn generate-sources
```
nicoloboschi, line 194: we have maven wrapper so you don't need to install your own maven

you can run 
```
./mvnw ...
```

nicoloboschi (changes requested):

README.md
```diff
@@ -199,3 +212,9 @@ or for all the services
 ```bash
 mini-langstream dev build
 ```
+If you run into issues with the build passing locally but failing the style checks in the CI, you can run:
+```bash
+mvn license:format
```
eolivelli, line 217: I prefer to suggest to use the standard Maven phase "generate-sources"
./mvn generate-sources

This way in  the future we can change the plugins without having to care about this docs

README.md
eolivelli: generate-sources is a phase that is implied by "install", you don't need the second command

eolivelli (commented):
Thanks for improving the developer experience.
I also left some minor comments

PR #722, opened 2023-11-16
Use Python asyncio gRPC

cbornet:
This has several advantages:
* better concurrency handling: Python is not good with threads (not OS threads)
* simplifies the code: the grpc-python async API is newer and better than the imperative one. And it makes it easier to handle methods that return Future as asyncio can wrap a Future.
* opens the path to supporting async read(), async write(), async process(), ... in agents. This will help to integrate with libs that have async APIs (eg. LangChain!)
* most importantly, will make the life easier to support TopicProducer::write that returns async.


nicoloboschi (approved):
Great work!

langstream-runtime/langstream-runtime-impl/src/main/python/Pipfile
```diff
@@ -6,7 +6,7 @@ fastavro = "*"
 "openai[embeddings]" = "*"
 "openai[datalib]" = "*"
 "openai[wandb]" = "*"
-langchain = "*"
+"langchain[openai]" = "*"
```
eolivelli, line 9: This change seems unrelated

eolivelli (approved):
Great stuff.
There is a change related to LangChain that seems unrelated.
I would push it to a separate PR

langstream-runtime/langstream-runtime-impl/src/main/python/Pipfile
cbornet: Done instead in https://github.com/LangStream/langstream/pull/723

PR #732, opened 2023-11-21
Fix Python agent read when an error is raised

cbornet:
The iterator method and context read/write APIs can't be mixed.
So this PR changes to use only the context API in read operation.
Also simplifies the tests by using the context API when it's easier than the iterator.


eolivelli (approved):
+1

Issue #649, opened 2023-10-25
Custom python agents: add support for writing to a topic


eolivelli:

eolivelli:
We don't need support for Consumers at the moment

cbornet:
done

PR #733, opened 2023-11-21
Add topic producer API Python side

cbornet:
See #649 

After that, remaining work for topic producer will be:
* Support AVRO schema
* Add e2e tests

### DIFF
--- a/langstream-runtime/langstream-runtime-impl/src/main/python/langstream_grpc/api.py
+++ b/langstream-runtime/langstream-runtime-impl/src/main/python/langstream_grpc/api.py
@@ -63,12 +63,28 @@ class Record(ABC):
 RecordType = Union[Record, dict, list, tuple]
 
 
+class TopicProducer(ABC):
+    """The topic producer interface"""
+
+    async def awrite(self, topic: str, record: Record):
+        """Write a record to a topic (for async methods)."""
+        pass
+
+    def write(self, topic: str, record: Record) -> Future:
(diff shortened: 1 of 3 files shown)


langstream-runtime/langstream-runtime-impl/src/main/python/langstream_grpc/api.py
```diff
@@ -63,12 +63,28 @@ def headers(self) -> List[Tuple[str, Any]]:
...
 
 
+class TopicProducer(ABC):
+    """The topic producer interface"""
+
+    async def write(self, topic: str, record: Record):
+        """Write a record to a topic (for async methods)."""
+        pass
+
+    def write_future(self, topic: str, record: Record) -> Future:
```
cbornet, line 73: Suggestions for a better name ?

eolivelli: what about "awrite" for the async write and "write" for the sync version ?

cbornet: Yes, I've thought of that. Now both methods are kinda "async", one with async/await, the other with Future.

cbornet: But let's use `awrite` as it's a defacto convention when 2 methods do the same thing and one does it on the event loop.

PR #740, opened 2023-11-30
[webcrawler-source] Expose the allow-non-html-contents parameter

eolivelli:
Summary:
- allow users to configure "allow-non-html-contents"


eolivelli:
I am not adding the property in the examples because otherwise people with older versions of LangStream won't be able to run the examples.
Usually we wait for a couple of releases

suleymanekmekci:
https://github.com/LangStream/langstream/pull/739/files#diff-9c087bc8c8ac85313ed8c78e9b51e51fa7c644c52ecdf114cf964baae998716d

maybe test needs to be updated too

PR #747, opened 2023-12-13
Add more tests and add back validation for some agents

eolivelli:
Summary:
- add tests for Python sources with big messages
- add fn:length function
- add validation for agents with nested collections (like "compute") (the integration tests failed due to a type in the compute agent, but the validation was not working)
- suppress errors in some shutdown cases for Python agents (they hid the errors in the compute step while I was working on integration tests)
- fix the validation of EL expressions in the configuration

### DIFF
--- a/langstream-e2e-tests/src/test/java/ai/langstream/tests/PythonAgentsIT.java
+++ b/langstream-e2e-tests/src/test/java/ai/langstream/tests/PythonAgentsIT.java
@@ -24,7 +24,6 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.testcontainers.shaded.com.google.common.base.Strings;
 
 @Slf4j
 @ExtendWith(BaseEndToEndTest.class)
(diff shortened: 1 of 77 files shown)


langstream-e2e-tests/src/test/java/ai/langstream/tests/PythonAgentsIT.java
```diff
@@ -24,6 +24,7 @@
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.testcontainers.shaded.com.google.common.base.Strings;
```
cbornet, line 27: Should probably not use a shaded dep.

langstream-e2e-tests/src/test/java/ai/langstream/tests/PythonAgentsIT.java
```diff
@@ -115,8 +116,12 @@ public void testSource() {
                                 .formatted(applicationId)
                                 .split(" "));
         log.info("Output: {}", output);
+        String bigPayload = Strings.repeat("test", 10000);
```
cbornet, line 119:
```suggestion
        String bigPayload = "test".repeat(10000);
```

eolivelli: good catch. fixed

nicoloboschi (approved):

PR #755, opened 2024-01-21
Add an S3 processor agent

cdbartholomew:
This agent will read a single file from an S3 bucket. The name of the file to read is taken from the record.


nicoloboschi (commented):
LGTM

langstream-agents/langstream-agent-s3/src/main/java/ai/langstream/agents/s3/S3Processor.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    private void processRecord(Record record, RecordSink recordSink) {
+        log.debug("Processing record {}", record.toString());
+        MutableRecord context = MutableRecord.recordToMutableRecord(record, true);
+        final JsonRecord jsonRecord = context.toJsonRecord();
+
+        String fileName = objectTemplate.execute(jsonRecord);
+
+        log.info("Processing file {}", fileName);
```
eolivelli, line 93: we should define what happens in case fileName is null or empty, maybe we  could "skip" the record ? (emit an empty list)

langstream-agents/langstream-agent-s3/src/main/java/ai/langstream/agents/s3/S3Processor.java
```diff
@@ -0,0 +1,194 @@
...
+            recordSink.emit(new SourceRecordAndResult(record, null, e));
+        }
+    }
+
+    @Override
+    protected Map<String, Object> buildAdditionalInfo() {
+        return Map.of("bucketName", bucketName);
+    }
+
+    private static class S3SourceRecord implements Record {
```
eolivelli, line 127: we could:
1) rename this to S3ProcessorRecord
2) use the same class of the S3Source
3) switch to SimpleRecord

langstream-agents/langstream-agent-s3/src/main/java/ai/langstream/agents/s3/S3Processor.java
```diff
@@ -0,0 +1,194 @@
...
+    public void init(Map<String, Object> configuration) throws Exception {
+        bucketName = configuration.getOrDefault("bucketName", "langstream-source").toString();
+        String endpoint =
+                configuration
+                        .getOrDefault("endpoint", "http://minio-endpoint.-not-set:9090")
+                        .toString();
+        String username = configuration.getOrDefault("access-key", "minioadmin").toString();
+        String password = configuration.getOrDefault("secret-key", "minioadmin").toString();
+        String region = configuration.getOrDefault("region", "").toString();
+        String objectName = configuration.getOrDefault("objectName", "").toString();
```
eolivelli, line 55: what about providing a sensible default, like "{{value.objectname}}" ?
ths way people don't have to bother with mustache for simple cases

langstream-runtime/langstream-runtime-impl/src/main/python/pyproject.toml
```diff
@@ -66,7 +66,7 @@ fastavro = "^1.9.2"
 optional = true
 
 [tool.poetry.group.full.dependencies]
-openai = {extras = ["datalib"], version = "^1.6.1"}
+openai = {extras = ["datalib"], version = "^1.10.0"}
```
eolivelli, line 69: this seems unrelated

eolivelli (changes requested):
Overall LGTM
but I have left some minor feedback.
I think we are on our way, this agent will be very useful

Looking forward to a PR to the docs website after we commit this patch

Thanks

langstream-api/src/main/java/ai/langstream/api/model/ResourcesSpec.java
```diff
@@ -16,21 +16,51 @@
...
+        if (value instanceof String) {
+            String stringValue = (String) value;
+            if (stringValue.startsWith("${") && stringValue.endsWith("}")) {
+                String variableName = stringValue.substring(2, stringValue.length() - 1);
+                return variableMap.getOrDefault(variableName, null);
+            }
+            try {
+                return Integer.parseInt(stringValue);
+            } catch (NumberFormatException e) {
+                System.err.println("Error parsing integer: " + e.getMessage());
```
eolivelli, line 58: logger ?

eolivelli (commented):
I see that this patch is accumulating changes not strictly related to the S3 Processor.
We can keep them but I would prefer to finalise the patch in this form and then follow up with other  improvements

eolivelli:
@cdbartholomew I suggest to closed this PR, this is more a diff between this repo and the main branch of another repo.

I would prefer to merge your changes using fewer smaller patches instead of this big PR.

I will be very happy to review the single patches

cdbartholomew:
Closing for now. Will update an open a new PR.

PR #756, opened 2024-01-26
docker run: prevent concurrent unpacking of nar files

eolivelli:


langstream-runtime/langstream-runtime-impl/src/main/java/ai/langstream/runtime/agent/AgentRunner.java
```diff
@@ -160,16 +160,19 @@ public void run(
...
-                sharedNarFileHandler != null
+                sharingNarFileHandler
                         ? sharedNarFileHandler
                         : new NarFileHandler(
                                 agentsDirectory,
                                 customLibClasspath,
                                 Thread.currentThread().getContextClassLoader());
         try {
-            narFileHandler.scan();
+            if (!sharingNarFileHandler) {
```
nicoloboschi, line 173: so who's going to call the scan method ?

eolivelli: it is LocalApplicationRunner

https://github.com/LangStream/langstream/blob/87f19a2c1c8e1a5a6b802fd5fe49a432b147ccd3/langstream-runtime/langstream-runtime-tester/src/main/java/ai/langstream/runtime/tester/LocalApplicationRunner.java#L94

